PR #143, opened 2015-07-24
Fix bug: svg element's className is not string.

armslave00:
Fix bug: svg element's className is not string. If mousemove to svg element, will throw an error.

### DIFF
--- a/src/Handler.js
+++ b/src/Handler.js
@@ -42,7 +42,8 @@ define(
                           || event.srcElement
                           || event.target;
 
-            return target && target.getAttribute("class").match(config.elementClassName)
+            var className = target && target.getAttribute("class");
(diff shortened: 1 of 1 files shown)


src/Handler.js
```diff
@@ -42,7 +42,8 @@ define(
                           || event.srcElement
                           || event.target;
 
-            return target && target.className.match(config.elementClassName)
+            var className = target && target.getAttribute("class");
```
Ovilia:
```suggestion
            var className = target && target.getAttribute('class');
```

Ovilia: Please use `'` instead of `"` for string.

Ovilia:
I found this has already out-of-date. Close but thanks for your contribution.

PR #160, opened 2015-12-05
fix path error

luqin:
替换zrender文件路径为相对路径:
'zrender/core/vector' => './vector'
'zrender/core/matrix' => './matrix'


pissang:
thx

PR #176, opened 2016-04-29
Change isArray condition to avoid clone object fails

xfanwu:
"instanceof Array" will return false and causing setOption() failed

### DIFF
--- a/src/core/util.js
+++ b/src/core/util.js
@@ -28,7 +28,7 @@ define(function(require) {
     function clone(source) {
         if (typeof source == 'object' && source !== null) {
             var result = source;
-            if (source instanceof Array) {
+            if (Object.prototype.toString.call(source) === '[object Array]') {
                 result = [];
                 for (var i = 0, len = source.length; i < len; i++) {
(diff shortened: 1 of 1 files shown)


pissang:
Can you provide your failed case?

xfanwu:
Sorry I forgot to mention that I am using Echart, which depend on zrender package. The setOption api is provided by Echart.

The instanceof operator will return false if we are in a multi-frame DOM environment.

See http://perfectionkills.com/instanceof-considered-harmful-or-how-to-write-a-robust-isarray/

Regards,
On April 29, 2016 at 10:44:48, Yi Shen (notifications@github.com) wrote:

Can you provide your failed case?

xfanwu:
Hello, could you fix this bug? My project has an Echart on top frame and I shoud pass data that constructed under child frame to top for setting new option.

You could try this:

```javascript
// under child frame environment: 
var arrayChildFrame = [1, 2, 3];
parent.arrayTopFrame = arrayChildFrame;

// then go to top frame environment:
arrayTopFrame instanceof Array // will return false
```

PR #294, opened 2017-11-12
Fix: use opts.width & opts.height in SVG Painter

chriswong:


src/svg/Painter.js
```diff
@@ -4,6 +4,7 @@
  */
 
 import {createElement} from './core';
+import * as util from './core/util';
```
Ovilia: This should be `import * as util from '../core/util';`

Ovilia:
@chriswong I just merged and fixed for you. :)

PR #322, opened 2018-02-02
Fix for "Passive event listeners" warning in console logs

acierto:
This is a fix for this annoying warnings, when you scroll the page:

screen shot 2018-02-02 at 11 29 22

There is a quite good explanation why it happens and how to prevent it.

My PR just contains the fix based on the recommendations.

Btw, this issue is already opened 

https://github.com/ecomfe/zrender/issues/238


acierto:
Anybody ever review PRs here? :)

Ovilia:
Hi @acierto , thanks for your commit. Please revert the files in dist since it should only be uploaded when release new version.

acierto:
Sure, reverted. Sorry didn't know your politics about dist.

It's done!

Ovilia:
It seems that passive event is a feature of Chrome, which I guess has nothing to do with `el.attachEvent`, which is used for IE.

acierto:
Yes true, does it hurt? As I know this issue is happening only on Chrome, but it's not a reason to ignore it :)

Ovilia:
@acierto Merged now. Thanks for your contribution!

acierto:
Thank you Olivia for accepting it!

Ovilia:
@acierto Normally we have monthly release, but since we are about to have Chinese New Year, maybe it will take some more time.

100pah:
@acierto Sorry I revert the passive support in 0397201


Because in zrender the dom event listeners delegate all of the upper events (zrender defined event) of elements. Some of those events need
`preventDefault`. For example, the zoom and pinch feature of echarts. That simply set the `mousewheel` event listener passive will disable those features.

We have not yet found a neat way to using passive. Before passive can be adopted, these issues should be considered:
(1) Whether and how a zrender user specifies an event listener passive. And by default, passive or not.
(2) How to tread that some zrender event listener is passive, and some is not. If we use other way but not preventDefault of `mousewheel` and `touchmove`, browser
compatibility should be handled.

And if you have been bothered by the 
Reproduct the console warning: 
```
[Violation] Added non-passive event listener to a scroll-blocking <some> event. Consider marking event handler as 'passive' to make the page more responsive.
```
Could you set console log level `verbose` in chrome dev tool to avoid the log displayed?

PowerSheetAI:
Can you enable passive listening just for the mousewheel and wheel events alone to both avoid many of these console log warnings and improve performance?

PROPOSED FIX (2 lined modified) without affecting preventDefault() use for other events:
```javascript
//*** ADD BELOW LINE: ***
const passiveEventNames = ['mousewheel', 'wheel']; //Used to prevent warning and performance hit: "[Violation] Added non-passive event listener to a scroll-blocking 'wheel' event. Consider marking event handler as 'passive' to make the page more responsive. See https://www.chromestatus.com/feature/5745543795965952". TO-OPTIMIZE: Instead use a whitelist instead for only events that need event.preventDefault() to further improve performance.

function mountLocalDOMEventListeners(instance, scope) {
    var domHandlers = scope.domHandlers;    
	if (env.pointerEventsSupported) {
        zrUtil.each(localNativeListenerNames.pointer, function (nativeEventName) {
            mountSingleDOMEventListener(scope, nativeEventName, function (event) {
                domHandlers[nativeEventName].call(instance, event);
            }, {passive: passiveEventNames.indexOf(nativeEventName) != -1}); //*** MODIFY THIS LINE ***
        });
```

Issue #339, opened 2018-04-27
Memory leak due to non-existing image rendering


kelvinau:
I actually found this issue when I was using EChart. The issue becomes very serious when I have a logic that tries to get image based on a variable. Size of `globalImageCache` keeps increasing after EChart's setOption.

When image is being loaded, `pendingWrap` from `createOrUpdateImage` is pushed to `pending` in `globalImageCache`. 
https://github.com/ecomfe/zrender/blob/fdd600891646644d215fcae015d3ebeb71e695f3/src/graphic/helper/image.js#L50
The `pending` array is only cleared when the image is loaded successfully. For non-existing image, i.e. onerror, the array will not be cleared.

Memory Increase:
![2018-04-27 00_26_47-image](https://user-images.githubusercontent.com/5885902/39350023-cddceddc-49b1-11e8-8fd7-3a321b42845e.png)

After I added a non-existing image 10 times
![2018-04-27 00_30_53-image](https://user-images.githubusercontent.com/5885902/39350202-45c2e75c-49b2-11e8-988d-e1816868152c.png)

After fix:
![2018-04-27 00_32_58-image](https://user-images.githubusercontent.com/5885902/39350283-90db15de-49b2-11e8-9485-f4b08e14f9dc.png)

JSFiddle Demo:
https://jsfiddle.net/kelvinau/y3axsx0w/

kelvinau:
Submitted pull request https://github.com/ecomfe/zrender/pull/340

PR #340, opened 2018-04-27
Fix memory leak due to non-existing image rendering

kelvinau:
Handle non-existing image by clearing `pending` in `globalImageCache` for image `onerror`
See Issue https://github.com/ecomfe/zrender/issues/339


kelvinau:
Will this PR be merged into master soon? This fix solves the critical memory problem to ECharts usage and our product. @pissang @Ovilia

pissang:
Thanks!! @kelvinau

PR #372, opened 2018-09-12
fix: ellipsis

yuqianma:


Ovilia:
We are not going to use `…` instead of `...`.

yuqianma:
Hi,
But putting the character aside, there's still a bug.

I made a separate pr: https://github.com/ecomfe/zrender/pull/384

PR #399, opened 2018-12-19
fix: mistake when drawing arc in PathProxy

Cmen117696:


Ovilia:
@Cmen117696 Thanks for your contribution! 🌟

PR #416, opened 2019-01-12
bug fixed: style.opacity not working for richText

Cmen117696:


src/contain/text.js
```diff
@@ -88,7 +88,7 @@ function getPlainTextRect(text, font, textAlign, textVerticalAlign, textPadding,
     return rect;
 }
 
-function getRichTextRect(text, font, textAlign, textVerticalAlign, textPadding, textLineHeight, rich, truncate) {
+function getRichTextRect(text, font, textAlign, textVerticalAlign, textPadding, rich, textLineHeight, truncate) {
```
Cmen117696: Since we need to pass so many parameters,
 I think it's better to use destructuring assignment syntax in es6, just like this.
```js
function getRichTextRect({text, font, textAlign, textVerticalAlign, textPadding, rich, textLineHeight, truncate}) {
     // main body
}
```
then we do not need to consider the order of parameters.

src/graphic/Text.js
```diff
@@ -67,6 +67,7 @@ Text.prototype = {
                 style.textAlign,
                 style.textVerticalAlign,
                 style.textPadding,
+                textContain.getLineHeight(style.font),
```
Ovilia: How does line height to do with opacity? If they're not related, you should probably use another PR for it.

PR #419, opened 2019-01-17
fixed color tool bug

troy351:


Ovilia:
Thanks for your contribution!
Can you describe what problem this bug cause?

troy351:
It's just a typo here since `color` is a string and `color[i] < 0` means nothing. should use `colorArr[i] < 0` instead.
As for the `if (!leftColor || !rightColor) return;` part, just do the `Can not read property xxx of null` check

Ovilia (dismissed):
LGTM. 

@100pah Please have a look and have it merged if it's OK.

PR #443, opened 2019-03-06
chore: add package lock and fix audit

Ovilia:
- Add `package-lock.json`
- Used `npm audit fix` to upgrade some npm package versions
- Changed license name according to https://spdx.org/licenses/


Jivanvl:
I was about to open an issue about this, thanks a lot!

PR #448, opened 2019-03-19
fix: bind mouse events in web worker

CarterLi:
Web workers do support addEventListener ( aka. EventTarget ) and `requestAnimationFrame`, but worker do not have the global object window.

Instead of window, we should use a better option to get the global object. [globalThis](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis) is the best option indeed, but we can't use it due to its poor compatibility. Browser and worker both support [self](https://developer.mozilla.org/en-US/docs/Web/API/Window/self), and it has very good browser compatibility. Using `self` can be a better option than using window.

This PR generally enables [OffscreenCanvas](https://developer.mozilla.org/en-US/docs/Web/API/OffscreenCanvas) support of zrender, which may greatly improve the performance of echarts.

Ref: https://github.com/apache/incubator-echarts/issues/9232

### DIFF
--- a/src/animation/requestAnimationFrame.js
+++ b/src/animation/requestAnimationFrame.js
@@ -1,12 +1,12 @@
 
 export default (
-    typeof window !== 'undefined'
+    typeof self !== 'undefined'
     && (
         // https://github.com/ecomfe/zrender/issues/189#issuecomment-224919809
-        (window.msRequestAnimationFrame && window.msRequestAnimationFrame.bind(window))
-        || window.requestAnimationFrame
-        || window.mozRequestAnimationFrame
-        || window.webkitRequestAnimationFrame
+        (self.msRequestAnimationFrame && self.msRequestAnimationFrame.bind(window))
+        || self.requestAnimationFrame
+        || self.mozRequestAnimationFrame
+        || self.webkitRequestAnimationFrame
     )
 ) || function (func) {
     setTimeout(func, 16);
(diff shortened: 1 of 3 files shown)


yuqianma:
Good catch!
But maybe it's unnecessary to instance a `handlerProxy`?
I had bypassed the proxy and used the handler directly in a different case.

CarterLi:
> I had bypassed the proxy and used the handler directly in a different case.

I'm not sure I understand what you mean, but do you mean using zrender.handler directly?

I know I can access `zrender.handler` by using `echartsInstance._zr.handler`, but I don't know if it's a good idea of accessing the private `_zr` property. But AFAIK trigger mouse events using `dispatchEvent` is a much easier way to reuse existing code.

yuqianma:
Got it. It's reasonable in echarts.
I'm using zrender alone. Considering some non-browser environments, I keep dom event things away to make it easier.

CarterLi:
I have tried bypassing `handlerProxy` by calling `echartsInstance._zr.zrender.trigger` directly (zrX, zrY and friends are set before), and got various errors.

I gave up,  and reverted back to `dispatchEvent` on OffscreenCanvas.

CarterLi:
It has been more than 1 year, any chance to review this?

pissang:
@CarterLi Sorry for missing this PR.

`addEventListener` in web worker is for receiving messages. It doesn't have the function of listening to mouse events. Ignoring it and still treat it as in the `dom` environment may lead to other problems in the further.

About mouse event. Perhaps you can reference to the code in ECharts GL. 
https://github.com/ecomfe/echarts-gl/blob/master/src/util/EChartsSurface.js#L140

It does a similar thing.

CarterLi:
Thanks @pissang

PR #458, opened 2019-04-22
Fix rich text align won't work in some case

susiwen8:
This PR relate to issue https://github.com/apache/incubator-echarts/issues/10327 in incubator-echarts


susiwen8:
Hi @100pah Could you review this PR?

Ovilia:
Thanks @susiwen8 . We will have a look at it.

PR #461, opened 2019-04-28
support swan

junmer:


src/Painter.js
```diff
@@ -477,8 +477,8 @@ Painter.prototype = {
             ctx.restore();
         }
 
-        if (env.wxa) {
-            // Flush for weixin application
+        if (env.miniApp) {
```
100pah: wxa should not be removed. Because env.wxa has been exported to developers.
and Baidu swan app is the different environment from wechat mini app.
If we intend to support Baidu swan app, probably we can add a new mark on env, 
like 'env.swan' or 'env.baiduSwan'  or 'env.bdswan' or something.

What's your opinion? @Ovilia

src/core/env.js
```diff
@@ -8,13 +8,28 @@
...
+/**
+ * get mini app api
+ *
+ * @see https://smartprogram.baidu.com/docs/develop/api/device_sys/#swan-getSystemInfoSync/
+ * @see https://developers.weixin.qq.com/miniprogram/dev/api/wx.getSystemInfoSync.html
+ * @return {boolean} result
+ */
+function getMiniApp() {
+    var wx = wx || null;
+    var swan = swan || null;
```
100pah: The global variable `wx` exists not only on wechat miniapp, but also on the other mini h5 pages, if they use the the js library provided by wechat.

src/core/env.js
```diff
@@ -8,18 +8,23 @@
...
 var env = {};
 
+var miniAppEnv = {
+    browser: {},
+    os: {},
+    node: false,
+    canvasSupported: true,
+    svgSupported: false,
+    touchEventsSupported: true,
+    domSupported: false
```
Ovilia: It would be better if you add `wxa: false, swan: false` here for default values.

Ovilia:
@junmer Thanks for your contribution. And I wonder if you have tested in Baidu swan application yet?

src/core/env.js
junmer: Reference the detect of browser . `env.firefox` has not default values . so do `env.wxa`.

junmer:
> @junmer Thanks for your contribution. And I wonder if you have tested in Baidu swan application yet?



> Yep. It works

PR #478, opened 2019-06-17
fix processGesture arguments error

soakit:
image


soakit:
removed in #480 
@Ovilia

PR #505, opened 2019-09-24
fix(svg): arc large bug #496

### LINKED ISSUE #496: arc( startAngle=0, endAngle>360) with svg render is not a circle
```js
var zr = zrender.init(document.getElementById('main'),{
            renderer: 'svg'
        });
        var arc = new zrender.Arc({
            shape: {
                cx: 200,
                cy: 200,
                r: 100,
                startAngle: 0,
                endAngle: 360,
            },
            style: {
                stroke: 'blue',
            }
        })
        zr.add(arc)
```

It works well  with canvas renderer,

Ovilia:
It fixes the bug with arc `large` value whose `endAngle` is greater than `Math.PI * 2`. It fixes and closes #496.

---

Canvas result of `text/svg-arc.html` (SVG is expected to be the same): 
canvas
The number in the circle represents the `(i, j)` value, and the circles are of `startAngle` to be `Math.PI / 4 * i` and `endAngles` to be `Math.PI / 4 * j`.

Before this fix, the SVG result is (wrong parts are highlighted with blue):
before-svg
This is mainly because the `large` calculation was wrong with `endAngle - startAngle` larger than `Math.PI * 2`.

---

After this fix, the SVG result is (difference with the Canvas result is in blue):
svg-after

I think currently the SVG result is correct, although we cannot say it's a bug with Canvas. Maybe we should adjust the behavior of Canvas cases.

[The Canvas spec](https://html.spec.whatwg.org/multipage/canvas.html#ellipse-method-steps) doesn't have a specific definition about this case and I tested Chrome and Firefox and they have different implementation about it. I created a [demo on jsfiddle](https://jsfiddle.net/ovilia/hzjgfq0e/14/) with circles of `endAngle - startAngle` to be `Math.PI * 2`.

Chrome renders 3PI to 5PI to be a circle, while Firefox renders an empty result.

I'd suggest merging this PR as the SVG results are the same as expected. As about Canvas, we could consider drawing a circle for all these cases.

### DIFF
--- a/src/svg/graphic.js
+++ b/src/svg/graphic.js
@@ -142,18 +142,20 @@ function pathDataToString(path) {
 
                 var dThetaPositive = Math.abs(dTheta);
                 var isCircle = isAroundZero(dThetaPositive - PI2)
-                    && !isAroundZero(dThetaPositive);
+                    || (clockwise ? dTheta >= PI2 : -dTheta >= PI2);
+
+                // Mapping to 0~2PI
+                var unifiedTheta = dTheta > 0 ? dTheta % PI2 : (dTheta % PI2 + PI2);
 
                 var large = false;
-                if (dThetaPositive >= PI2) {
+                if (isCircle) {
                     large = true;
                 }
                 else if (isAroundZero(dThetaPositive)) {
                     large = false;
                 }
                 else {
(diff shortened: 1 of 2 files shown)


pissang:
Very detailed test case!

PR #509, opened 2019-09-29
Add feat for minBI

deqingli:
+ fix code annotation

+ export getBoxPosition()

### DIFF
--- a/src/graphic/Displayable.js
+++ b/src/graphic/Displayable.js
@@ -1,5 +1,4 @@
 /**
- * 可绘制的图形基类
  * Base class of all displayable graphic objects
  * @module zrender/graphic/Displayable
  */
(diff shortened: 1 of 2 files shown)


src/graphic/Displayable.js
```diff
@@ -53,7 +53,7 @@ Displayable.prototype = {
     type: 'displayable',
 
     /**
-     * Displayable 是否为脏，Painter 中会根据该标记判断是否需要是否需要重新绘制
+     * Displayable 是否为脏，Painter 中会根据该标记判断是否需要重新绘制
```
100pah: Should better change to English

src/graphic/Path.js
```diff
@@ -177,6 +177,26 @@ Path.prototype = {
         this.path = new PathProxy();
     },
 
+    getBoundingRectWithoutStroke: function () {
+        var rect = this._rect;
```
100pah: Should not repeat the code. They are the same with `getBoundingRect`.

src/graphic/Displayable.js
deqingli: OK

PR #513, opened 2019-10-16
feat: add a callback for fixClipWithShadow

Ovilia:
So that to support it for Sausage shape defined in apache/incubator-echarts#11393.


Ovilia:
`fixClipWithShadow` is

PR #525, opened 2019-11-01
fix(svg): bug with brush in bar.html since 13d13a86

Ovilia:
fix(svg): bug with brush in bar.html since 13d13a86


pissang:
Fixed in https://github.com/ecomfe/zrender/pull/527

PR #526, opened 2019-11-01
Fix shadow display bug on SVG.

pissang:
There are two fixes in this pull request

1. Shadow is affected on the text, not main element. Like the shot below
MacHi 2019-11-01 21-04-52

This bug is caused by https://github.com/ecomfe/zrender/commit/409d52619f8a68807658b00cc53d91c6ce63de4d#diff-4bf718158d3bfa56f864a33987f9882aL238

shadow should be added on the `svgElement` instead of `prevSvgElement`, which can be the text element.

2. DOM of shadow is stored in the `style` property of SVG. And this `style` property can be replaced during the update of echarts. It will cause  this information lost and can't remove the `filter` attribute successfully in https://github.com/ecomfe/zrender/blob/master/src/svg/helper/ShadowManager.js#L123

So I fixed it by putting the DOM in `displayable` insteadof `style`.


pissang:
Closed. I should merge to release branch

PR #535, opened 2019-11-21
WIP(svg): clear text when mouse leaves

Ovilia:
Fix https://github.com/apache/incubator-echarts/issues/11543 .

This commit can remove the text. But when the mouse hovers on the circle again, the text doesn't show up.

### DIFF
--- a/src/svg/Painter.js
+++ b/src/svg/Painter.js
@@ -243,8 +243,8 @@ SVGPainter.prototype = {
             else if (!item.removed) {
                 for (var k = 0; k < item.count; k++) {
                     var displayable = newVisibleList[item.indices[k]];
-                    prevSvgElement = getTextSvgElement(displayable)
-                        || getSvgElement(displayable) || prevSvgElement;
+                    var svgElement = getSvgElement(displayable);
+                    var textSvgElement = getTextSvgElement(displayable);
 
                     var svgElement = getSvgElement(displayable);
                     var textSvgElement = getTextSvgElement(displayable);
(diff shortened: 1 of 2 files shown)


pissang:
The `textSvgEl` is inserted to the document in https://github.com/ecomfe/zrender/blob/master/src/svg/Painter.js#L231 . And it will only be executed when element is being paint in the first time.

So if you remove it from the document when `text` property is set to null in https://github.com/ecomfe/zrender/pull/535/files#diff-5bb01bd8925bd801f17e33bdb859d019R515. It won't be insert to the document again when `text` is set to another value.

A better way is to hide this text element by removing the `tspan` elements.

Ovilia:
@pissang But it has the same behavior with `tspan`:

```js
function removeOldTextNode(el) {
    if (el && el.__textSvgEl) {
        var children = el.__textSvgEl.children;
        for (var i = 0; i < children.length; ++i) {
            if (children[i]) {
                el.__textSvgEl.removeChild(children[i]);
            }
        }
    }
}
```

pissang:
It seems you need to reset more cached properties.

```js

function removeOldTextNode(el) {
    if (el && el.__textSvgEl) {
        el.__textSvgEl.parentNode.removeChild(el.__textSvgEl);
        el.__textSvgEl = null;
        el.__tspanList = [];
        el.__text = null;
    }
}
```

Also the new created svg text dom needs to be add to document when updating the element in 
https://github.com/ecomfe/zrender/blob/master/src/svg/Painter.js#L244
```js
else if (!item.removed) {
                for (var k = 0; k < item.count; k++) {
                    var displayable = newVisibleList[item.indices[k]];

                    var svgElement = getSvgElement(displayable);
                    var textSvgElement = getTextSvgElement(displayable);

                    this.gradientManager.markUsed(displayable);
                    this.gradientManager
                        .addWithoutUpdate(svgElement || textSvgElement, displayable);

                    this.shadowManager.markUsed(displayable);
                    this.shadowManager
                        .addWithoutUpdate(svgElement || textSvgElement, displayable);

                    this.clipPathManager.markUsed(displayable);

                    if (textSvgElement) {   // Insert text.
                        insertAfter(svgRoot, textSvgElement, svgElement);
                    }
                    prevSvgElement = svgElement
                        || textSvgElement || prevSvgElement;

                }
```

PR #539, opened 2019-11-27
Add highlight and downplay on rect el

deqingli:


src/graphic/shape/Rect.js
```diff
@@ -58,5 +58,13 @@ export default Path.extend({
         }
         ctx.closePath();
         return;
+    },
+
+    highlight: function () {
```
100pah: There is no the concept of highlight/downplay/emphasis/normal in zrender yet.
I think it is not necessary to introduce those concept to zrender.

PR #568, opened 2020-02-13
Fix 11781: save svg image with toolbox

Ovilia:
See https://github.com/apache/incubator-echarts/pull/12144 for details.

### DIFF
--- a/src/svg/Painter.js
+++ b/src/svg/Painter.js
@@ -134,6 +134,14 @@ SVGPainter.prototype = {
         return this._viewport;
     },
 
+    getSvgDom: function () {
+        return this._svgDom;
+    },
+
+    getSvgRoot: function () {
+        return this._svgRoot;
+    },
+
     getViewportRootOffset: function () {
         var viewportRoot = this.getViewportRoot();
         if (viewportRoot) {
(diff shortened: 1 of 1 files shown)


src/svg/Painter.js
```diff
@@ -146,7 +161,17 @@ SVGPainter.prototype = {
...
+        // Otherwise, the exported SVG don't have background.
+        var bgNode = createElement('rect');
+        bgNode.setAttribute('width', this.getWidth());
+        bgNode.setAttribute('height', this.getHeight());
+        bgNode.setAttribute('x', 0);
+        bgNode.setAttribute('y', 0);
+        bgNode.setAttribute('id', 0);
+        bgNode.style.fill = backgroundColor;
+        this._backgroundRoot.appendChild(bgNode);
+        this._backgroundNode = bgNode;
```
pissang: Need to remove the previous created `_backgroundNode`?

src/svg/Painter.js
```diff
@@ -394,7 +427,7 @@ SVGPainter.prototype = {
 
     pathToDataUrl: function () {
```
pissang: Perhaps it's better to be named `toDataUrl`? It can be a common method for all implemented painters(canvas / svg)

Also `path` is a term that being used as a type of graphic element. It may be confusing here.

PR #617, opened 2020-07-18
Fix echarts datazoom: add 'keys' method to HashMap

susiwen8:
https://github.com/apache/incubator-echarts/blob/next/src/component/dataZoom/DataZoomModel.ts#L546

TS throw error that said: `DataZoomTargetAxisInfoMap ` don't have `keys` method


susiwen8:
@pissang @100pah Please review this PR

PR #618, opened 2020-07-20
feat: support for dash line scale. 

plainheart:
For apache/incubator-echarts#12961.

### DIFF
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -170,10 +170,12 @@ function brushPath(ctx: CanvasRenderingContext2D, el: Path, style: PathStyleProp
 
     if (lineDash) {
         const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
-        lineDash = map(lineDash, function (rawVal) {
-            return rawVal / lineScale;
-        });
-        lineDashOffset /= lineScale;
+        if (lineScale && lineScale !== 1) {
+            lineDash = map(lineDash, function (rawVal) {
+                return rawVal / lineScale;
+            });
+            lineDashOffset /= lineScale;
+        }
     }
 
     let needsRebuild = true;
@@ -316,12 +318,14 @@ function brushText(ctx: CanvasRenderingContext2D, el: TSpan, style: TSpanStylePr
             let lineDashOffset = style.lineDashOffset;
             if (lineDash) {
                 const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
-                lineDash = map(lineDash, function (rawVal) {
-                    return rawVal / lineScale;
-                });
-                lineDashOffset /= lineScale;
+                if (lineScale && lineScale !== 1) {
+                    lineDash = map(lineDash, function (rawVal) {
+                        return rawVal / lineScale;
+                    });
+                    lineDashOffset /= lineScale;
+                }
             }
-            ctx.setLineDash(lineDash as number[] || []);
+            ctx.setLineDash(lineDash || []);
             ctx.lineDashOffset = lineDashOffset;
         }
 
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -14,6 +14,7 @@ import TSpan, {TSpanStyleProps} from '../graphic/TSpan';
 import { DEFAULT_FONT } from '../contain/text';
 import { IncrementalDisplayable } from '../export';
 import { MatrixArray } from '../core/matrix';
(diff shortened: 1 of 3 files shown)


src/canvas/graphic.ts
```diff
@@ -151,15 +152,23 @@ function brushPath(ctx: CanvasRenderingContext2D, el: Path, inBatch: boolean) {
...
     const ctxLineDash = !!ctx.setLineDash;
 
     // Update path sx, sy
     const scale = el.getGlobalScale();
     path.setScale(scale[0], scale[1], el.segmentIgnoreThreshold);
 
+    if (lineDash) {
+        const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
+        lineDash = map(lineDash, function (rawVal) {
+            return rawVal / lineScale;
```
pissang, line 167: It's better to only been mapped when `lineScale` is not `1` to reduce the potential cost.

src/svg/graphic.ts
```diff
@@ -106,10 +107,14 @@ function bindStyle(svgEl: SVGElement, style: AllStyleOption, el?: Path | TSpan |
...
         attr(svgEl, 'paint-order', style.strokeFirst ? 'stroke' : 'fill');
         attr(svgEl, 'stroke-opacity', (style.strokeOpacity != null ? style.strokeOpacity * opacity : opacity) + '');
-        const lineDash = style.lineDash;
+        let lineDash = style.lineDash;
         if (lineDash) {
-            attr(svgEl, 'stroke-dasharray', (style.lineDash as number[]).join(','));
-            attr(svgEl, 'stroke-dashoffset', mathRound(style.lineDashOffset || 0) + '');
+            lineDash = map(lineDash, function (rawVal) {
+                return rawVal / strokeScale;
+            });
```
pissang, line 114: Similar to the above issue

src/canvas/graphic.ts
```diff
@@ -299,6 +308,19 @@ function brushText(ctx: CanvasRenderingContext2D, el: TSpan) {
...
         ctx.textAlign = style.textAlign;
         ctx.textBaseline = style.textBaseline;
+        if (ctx.setLineDash) {
+            let lineDash = style.lineDash;
+            let lineDashOffset = style.lineDashOffset;
+            if (lineDash) {
+                const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
+                lineDash = map(lineDash, function (rawVal) {
+                    return rawVal / lineScale;
+                });
```
pissang, line 318: Similar to the above issue

src/graphic/Text.ts
```diff
@@ -741,6 +747,8 @@ class ZRText extends Displayable<TextProps> {
 
         if (textStroke) {
             subElStyle.lineWidth = retrieve3(tokenStyle.lineWidth, style.lineWidth, defaultLineWidth);
+            subElStyle.lineDash = retrieve3(tokenStyle.lineDash, style.lineDash, []);
+            subElStyle.lineDashOffset = style.lineDashOffset || 0;
```
pissang, line 751: Perhaps the default `lineDash` should be `null`  instead of an empty array. Or in the `brush` method it will enter drawing `lineDash` branch https://github.com/ecomfe/zrender/pull/618/files#diff-0ac9120ba7869da49840beff4c68ab8eR315

plainheart:
Tweaked. Please help review later.

src/canvas/graphic.ts
```diff
@@ -158,15 +159,25 @@ function brushPath(ctx: CanvasRenderingContext2D, el: Path, style: PathStyleProp
...
 
     const ctxLineDash = !!ctx.setLineDash;
 
     // Update path sx, sy
     const scale = el.getGlobalScale();
     path.setScale(scale[0], scale[1], el.segmentIgnoreThreshold);
 
+    if (lineDash) {
+        const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
+        if (lineScale && lineScale !== 1) {
```
plainheart, line 173: @pissang Suddenly I remember a previous doubt about the value of `Transformable.getLineScale()`, that is, is it possible to be `0`? If possible, maybe we should set `lineDash` as an empty array and `lineDashOffset` as `0` to avoid the calculation?

plainheart: I found it's possible when the width is `0`. It won't have any effect even if we don't set `lineDash` as an empty array and `lineDashOffset` as 0 since this element is invisible for its stroke width is `0`. So, there may be no need to do it.

PR #620, opened 2020-07-26
Rich text padding

susiwen8:
This PR related to https://github.com/apache/incubator-echarts/pull/12947

Left padding in not working in rich mode

### DIFF
--- a/src/graphic/Text.ts
+++ b/src/graphic/Text.ts
@@ -606,11 +606,14 @@ class ZRText extends Displayable<TextProps> {
         let xLeft = adjustTextX(baseX, contentBlock.contentWidth, textAlign);
         let lineTop = adjustTextY(baseY, contentBlock.contentHeight, verticalAlign);
 
-        if (textPadding) {
+        if (textPadding && textAlign === 'left') {
             xLeft += textPadding[3];
         }
 
-        const xRight = xLeft + contentWidth;
+        let xRight = xLeft + contentWidth;
+        if (textPadding && textAlign === 'right') {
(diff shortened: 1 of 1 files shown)


src/graphic/Text.ts
```diff
@@ -606,6 +606,10 @@ class ZRText extends Displayable<TextProps> {
         let xLeft = adjustTextX(baseX, contentBlock.contentWidth, textAlign);
         let lineTop = adjustTextY(baseY, contentBlock.contentHeight, verticalAlign);
 
+        if (textPadding) {
+            xLeft += textPadding[3];
```
Ovilia, line 610: Should we consider `textAlign` before using adding? For example, if `textAlign` is `'center'`, maybe `xLeft` should not add the text padding?

susiwen8: Yes, you are right, and also consider that `textAlign` was `right`, `xRight` should subtract padding @Ovilia

pissang: It seems I made it wrong when refracting. We can roll it back to

https://github.com/ecomfe/zrender/blob/master/src/graphic/helper/text.js

```ts

        const boxX = adjustTextX(baseX, outerWidth, textAlign);
        const boxY = adjustTextY(baseY, outerHeight, verticalAlign);
        let xLeft = boxX;
        let lineTop = boxY;

        if (textPadding) {
            xLeft += textPadding[3];
        }

        let xRight = xLeft + innerWidth;

```

susiwen8: @pissang done, I have revert the code

PR #628, opened 2020-08-17
feat: normalize dash style in zrender.

plainheart:
Related to apache/incubator-echarts#12961

### DIFF
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -160,7 +160,7 @@ function brushPath(ctx: CanvasRenderingContext2D, el: Path, style: PathStyleProp
         }
     }
 
-    let lineDash = style.lineDash = normalizeLineDash(style.lineDash, style.lineWidth);
+    let lineDash = normalizeLineDash(style.lineDash, style.lineWidth);
     let lineDashOffset = style.lineDashOffset;
 
     const ctxLineDash = !!ctx.setLineDash;
@@ -315,7 +315,7 @@ function brushText(ctx: CanvasRenderingContext2D, el: TSpan, style: TSpanStylePr
         ctx.textAlign = style.textAlign;
         ctx.textBaseline = style.textBaseline;
         if (ctx.setLineDash) {
-            let lineDash = style.lineDash = normalizeLineDash(style.lineDash);
+            let lineDash = normalizeLineDash(style.lineDash);
             let lineDashOffset = style.lineDashOffset;
             if (lineDash) {
                 const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -315,7 +315,7 @@ function brushText(ctx: CanvasRenderingContext2D, el: TSpan, style: TSpanStylePr
         ctx.textAlign = style.textAlign;
         ctx.textBaseline = style.textBaseline;
         if (ctx.setLineDash) {
-            let lineDash = normalizeLineDash(style.lineDash);
+            let lineDash = style.lineDash && style.lineWidth > 0 && normalizeLineDash(style.lineDash, style.lineWidth);
             let lineDashOffset = style.lineDashOffset;
             if (lineDash) {
                 const lineScale = (style.strokeNoScale && el.getLineScale) ? el.getLineScale() : 1;
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -15,6 +15,7 @@ import { DEFAULT_FONT } from '../contain/text';
 import { IncrementalDisplayable } from '../export';
 import { MatrixArray } from '../core/matrix';
 import { map } from '../core/util';
+import { normalizeLineDash } from '../graphic/helper/dashStyle';
 
 const pathProxyForDraw = new PathProxy(true);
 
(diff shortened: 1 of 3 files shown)


src/canvas/graphic.ts
```diff
@@ -314,7 +315,7 @@ function brushText(ctx: CanvasRenderingContext2D, el: TSpan, style: TSpanStylePr
         ctx.textAlign = style.textAlign;
         ctx.textBaseline = style.textBaseline;
         if (ctx.setLineDash) {
-            let lineDash = style.lineDash;
+            let lineDash = style.lineDash = normalizeLineDash(style.lineDash);
             let lineDashOffset = style.lineDashOffset;
```
pissang, line 319: It's better not to modify the original value of `style.lineDash`

plainheart: Tweaked in https://github.com/ecomfe/zrender/pull/628/commits/cf5e31670d4ca914b6f24dc40071e3ad6d43e23f

src/canvas/graphic.ts
```diff
@@ -159,7 +160,7 @@ function brushPath(ctx: CanvasRenderingContext2D, el: Path, style: PathStyleProp
         }
     }
 
-    let lineDash = style.lineDash;
+    let lineDash = normalizeLineDash(style.lineDash, style.lineWidth);
```
pissang, line 163: Only normalize when `lineDash` exists. Or it will always enter https://github.com/ecomfe/zrender/blob/cf5e31670d4ca914b6f24dc40071e3ad6d43e23f/src/canvas/graphic.ts#L173 branch, which is an unnecessary extra cost.

src/graphic/helper/dashStyle.ts
```diff
@@ -0,0 +1,30 @@
...
+    if (dashArray) {
+        return dashArray;
+    }
+    lineWidth = lineWidth || 1;
+    return lineType === 'dashed' 
+        ? [4 * lineWidth, 2 * lineWidth]
+        : lineType === 'dotted' 
+            ? [lineWidth]
+            : [];
+}
```
pissang, line 30: Perhaps `normalizeDashArray` can be combined with this condition to simplify the check.

```ts
 return lineType === 'dashed' 
        ? [4 * lineWidth, 2 * lineWidth]
        : lineType === 'dotted' 
            ? [lineWidth]
            : isNumber(lineType)
                ? [lineType] : isArray(lineType) ? lineType : null;
```

Also, I think we should return `null` if `lineDash` is invalid to reduce the `setLineDash` call as more as possible.

plainheart: Thank you, I learned this from your guide and will be careful of it in the future.

src/graphic/helper/dashStyle.ts
```diff
@@ -0,0 +1,14 @@
+import { isArray, isNumber } from '../../core/util';
+
+export function normalizeLineDash(lineType: any, lineWidth?: number): number[] | false {
+    if (!lineType || lineType === 'solid' || (lineWidth != null && lineWidth <= 0)) {
+        return null;
```
pissang, line 5:
```ts
(lineWidth != null && lineWidth <= 0))
```
can be simplified to
```ts
!(lineWidth > 0)
```

It will also reject other potential invalid NaN values.

Sorry for being very critical in this review. Because it's at the very bottom of code that may affect a lot.

src/svg/graphic.ts
```diff
@@ -107,7 +108,7 @@ function bindStyle(svgEl: SVGElement, style: AllStyleOption, el?: Path | TSpan |
         // stroke then fill for text; fill then stroke for others
         attr(svgEl, 'paint-order', style.strokeFirst ? 'stroke' : 'fill');
         attr(svgEl, 'stroke-opacity', (style.strokeOpacity != null ? style.strokeOpacity * opacity : opacity) + '');
-        let lineDash = style.lineDash;
+        let lineDash = style.lineDash && normalizeLineDash(style.lineDash, el instanceof TSpan ? void 0 : style.lineWidth);
         if (lineDash) {
```
pissang, line 112: Sincde normalizeLineDash will reject `0` lineWidth.

```js
el instanceof TSpan ? void 0 : style.lineWidth
```

Can be 

```js
el instanceof TSpan ? 0 : style.lineWidth
```

plainheart: Here I want the text element to ignore the judgment `lineWidth <= 0` and continue to invoke the bottom logic in `normalizeLineDash` function. If we pass `0` to it, the returned `lineDash` will be `null`, which is unexpected.

plainheart: I will remove this condition.

plainheart:
I found there may be something wrong with canvas renderer. The legend symbol also have dash style. This is not unexpected.
I'm not sure how to fix them.

image

pissang:
I guess it's because the state set by `setLineDash` in `brushText` leak to the `brushPath`. 

In `brushPath` it will use `setLineDash([])` to restore the line dash state if it's changed in
https://github.com/ecomfe/zrender/blob/5b9003598c177c7cd4a5f4b491eb5524ae9f4351/src/canvas/graphic.ts#L247

It seems `brushText` doesn't do that.

plainheart:
I just made a commit and should I move the added logic to `if (text)` block?

pissang:
@plainheart Yeah I think it will be better

plainheart:
@pissang Done. Please help review it again.

pissang (approved):

PR #630, opened 2020-08-23
perf: don't slow down other modern browsers for IE compatibility

CarterLi:


pissang:
Hi, @CarterLi. Do you mind explaining why this change will have better performance in other browsers?

CarterLi:
`window.requestAnimationFrame.bind(window)` is a hack only for working around an issue of the old IE browsers, and should not affect other more commonly used browsers like Chrome.

All IE browsers ( and maybe non-blink-core edge ) including IE11 supports `msRequestAnimationFrame`, so only IE will use `window.msRequestAnimationFrame.bind(window)`, other browsers will only see `requestAnimationFrame`

The performance benefit of one function call is pretty small. But requestAnimationFrame will be called every frame. the benefit will be accumulate.

image

image

pissang:
Understand. Even though requestAnimationFrame itself is not a hot spot. I think it's still a good fix. Thanks!

pissang (approved):

PR #638, opened 2020-09-18
feat: decal

Ovilia:
See issue apache/incubator-echarts#13263 and ECharts PR apache/incubator-echarts#13304


src/core/WeakMap.ts
```diff
@@ -0,0 +1,38 @@
+let wmUniqueIndex = 0;
+
+export default class WeakMap<K extends object, V> {
+
+    protected _id: string;
+
+    constructor() {
+        this._id = '__\0ec_wm_' + wmUniqueIndex++ + '_' + Math.random().toFixed(5);
+    }
```
pissang, line 9: Id generating strategy can be same to https://github.com/apache/incubator-echarts/blob/next/src/util/model.ts#L721

src/core/WeakMap.ts
```diff
@@ -0,0 +1,38 @@
...
+
+    get(key: K): V {
+        return (this._guard(key) as any)[this._id];
+    }
+
+    set(key: K, value: V): WeakMap<K, V> {
+        (this._guard(key) as any)[this._id] = value;
```
pissang, line 16: It's better to use
```ts
defineProperty(key, { enumerable: false, configurable: false });
```
So this property won't be iterated when doing `clone` or something else. 

Also need to check if definePropert exists

pissang (changes requested):

src/graphic/helper/patternSize.ts
```diff
@@ -0,0 +1 @@
+function getPatternSize(pattern: PatternObject)
```
pissang, line 1: Should remove this unused file

src/svg/helper/PatternManager.ts
```diff
@@ -0,0 +1,258 @@
...
+                const scaleX = pattern.scaleX || 1;
+                const scaleY = pattern.scaleY || 1;
+                const transform = `translate(${x}, ${y}) rotate(${rotation}) scale(${scaleX}, ${scaleY})`;
+                patternDom.setAttribute('patternTransform', transform);
+
+                img.setAttribute('href', imageSrc);
+                img.setAttribute('x', '0');
+                img.setAttribute('y', '0');
+
+                getPatternSize(pattern, (width, height) => {
```
pissang, line 173: `createOrUpdateImage` can do the same thing.

```ts
createOrUpdateImage(imageSrc, img, { dirty(){} }, (image) => {
  patternDom.setAttribute('width', image.width);
  patternDom.setAttribute('height', image.height);
});
```

src/svg/helper/PatternManager.ts
```diff
@@ -0,0 +1,258 @@
...
+        if (size) {
+            if (typeof cb === 'function') {
+                cb(size.width, size.height);
+            }
+            return;
+        }
+    }
+    else if (pattern.image instanceof HTMLCanvasElement) {
+        src = pattern.image.toDataURL();
+    }
```
pissang, line 238: Can use canvas.width, canvas.height directly here. It's just a remind since `getPatternSize` can be removed and replaced with `createOrUpdateImage`.

pissang (changes requested):

src/graphic/Path.ts
```diff
@@ -151,6 +160,36 @@ class Path<Props extends PathProps = PathProps> extends Displayable<Props> {
...
+                = this._decalEl || new Path();
+            if (decalEl.buildPath === Path.prototype.buildPath) {
+                decalEl.buildPath = ctx => {
+                    this.buildPath(ctx, this.shape);
+                };
+            }
+
+            decalEl.silent = true;
+            decalEl.style.fill = style.fill ? style.decal : null;
+            decalEl.style.stroke = style.stroke ? style.stroke : null;
```
pissang, line 178: Should inherit `opacity`, `lineWidth` in decalEl

pissang (changes requested):

PR #643, opened 2020-09-30
feat(shape/sector): support cornnerRadius & innerCornerRadius.

plainheart:
For the PR on ECharts apache/incubator-echarts#13378

### DIFF
--- a/src/graphic/helper/roundSector.ts
+++ b/src/graphic/helper/roundSector.ts
@@ -1,0 +1,541 @@
+import PathProxy from '../../core/PathProxy';
+import { normalizeRadian } from '../../contain/util'
+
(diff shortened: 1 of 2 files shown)


pissang:
I'm not sure. But I suppose the code of calculating the round corner can be simplified with a neat formula. 

The code in d3-shape seems to be much simpler. https://github.com/d3/d3-shape/blob/master/src/arc.js#L116

plainheart:
Thanks a lot. That's neater and more simple indeed. I've simplified related logic in this commit https://github.com/ecomfe/zrender/pull/643/commits/7f66d8fe556475d3d39a5fdc709a57f584be7a65.

pissang (approved):

PR #645, opened 2020-10-15
fix(tool/color): enhance the color compat.

plainheart:
### Changes

(1) add fallback for `rgba(r, g, b)`. It will be converted to `rgba(r, g, b, 1)`.
(2) support 8-digit hex color(`#rrggbbaa`) which uses the last pair of digits as alpha, see https://drafts.csswg.org/css-color/#hex-notation and also support the form of `#rgba` (4-digit hex color in the form of `#rgba`).

### Results

(1) `rgba(17, 163, 69)` will be converted to `[17, 163, 69, 1]` **(Previous result is `undefined`)**
(2) `#14c4baff` will be converted to `[20, 196, 186, 1]` **(Previous result is `undefined`)**
(3) `#07f0` will be converted to `[0, 119, 255, 0]` **(Previous result is `undefined`)**

### Fixed issues

This fixes apache/incubator-echarts#13433 and enables the forms of color to be more flexible.

### Test cases
Please refer to `test/ut/spec/tool/color.test.ts`

### DIFF
--- a/src/tool/color.ts
+++ b/src/tool/color.ts
@@ -180,25 +180,28 @@ export function parse(colorStr: string, rgbaArr?: number[]): number[] {
         return rgbaArr;
     }
 
-    // #abc and #abc123 syntax.
+    // supports the forms #rgb, #rrggbb, #rgba, #rrggbbaa
+    // #rrggbbaa(use the last pair of digits as alpha)
+    // see https://drafts.csswg.org/css-color/#hex-notation
     if (str.charAt(0) === '#') {
-        if (str.length === 4) {
-            let iv = parseInt(str.substr(1), 16);  // TODO(deanm): Stricter parsing.
+        if (str.length === 4 || str.length === 5) {
+            const iv = parseInt(str.slice(1, 4), 16);  // TODO(deanm): Stricter parsing.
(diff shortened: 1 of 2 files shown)


src/tool/color.ts
```diff
@@ -180,25 +180,28 @@ export function parse(colorStr: string, rgbaArr?: number[]): number[] {
...
 
-    // #abc and #abc123 syntax.
+    // supports the forms #rgb, #rrggbb, #rgba, #rrggbbaa
+    // #rrggbbaa(use the last pair of digits as alpha)
+    // see https://drafts.csswg.org/css-color/#hex-notation
     if (str.charAt(0) === '#') {
-        if (str.length === 4) {
-            let iv = parseInt(str.substr(1), 16);  // TODO(deanm): Stricter parsing.
+        if (str.length === 4 || str.length === 5) {
+            const iv = parseInt(str.slice(1, 4), 16);  // TODO(deanm): Stricter parsing.
```
pissang, line 188: `str.length` can be cached in a variable since they are used more frequently now

plainheart: Tweaked in the commit a74bc35a2dea5966c81cf3c5a9e51359619ef662

pissang (approved):

PR #652, opened 2020-11-02
fix(sector): skip drawing corner radius if no arc.

plainheart:
When the corner radius is specified but the angle of sector is smaller than `1e-4`, we should not draw the corners.
Otherwise, we will get an unexpected result like the following screenshots.

image

image


plainheart:
@pissang Please help review this.

pissang:
LGTM

pissang (approved):

PR #673, opened 2020-12-01
fix(svg): fix chart cannot be exported with SVG renderer in IE.

plainheart:
**1) fix charts cannot be exported when using SVG renderer in IE.**
See apache/incubator-echarts#13732

****2)** fix exported SVG file cannot be opened when using external images.**
See apache/incubator-echarts#13655


src/svg/Painter.ts
```diff
@@ -458,7 +460,15 @@ class SVGPainter implements PainterBase {
 
     toDataURL() {
         this.refresh();
-        const html = encodeURIComponent(this._svgDom.outerHTML.replace(/></g, '>\n\r<'));
+        const outerHTML = this._svgDom.outerHTML
+            // outerHTML of `svg` tag is not supported in IE, use `parentNode.innerHTML` instead
+            // PENDING: Or use `new XMLSerializer().serializeToString(svg)`?
+            || (this._svgDom.parentNode && this._svgDom.parentNode as HTMLElement).innerHTML;
```
pissang, line 466: `this._svgDom` can be assigned to a temporal variable `svgDom` for a smaller size.

src/svg/Painter.ts
```diff
@@ -458,7 +460,15 @@ class SVGPainter implements PainterBase {
...
+        const html = encodeURIComponent(
+            outerHTML
+                .replace(/></g, '>\n\r<')
+                // PENDING: remove &quot; ?
+                .replace(/url\(&quot;([^)]*)&quot;\)/gm, 'url($1)'));
```
pissang, line 471: When will this HTML have `&quot; `?

plainheart: It's only in IE. Some quotation marks will be encoded and included in exported SVG files.
```html
<g clip-path="url(&quot;#zr2323-clip-0&quot;)">...</g>
```
But in fact, it can be opened and display correctly even if it has `&quot;`. So I'm not sure if they should be removed.

pissang: I think we can keep it if it has no issue. Or we only do it in IE to reduce the possibility of bringing unexpected problems.

plainheart: OK. Removed it.

pissang (approved):

PR #686, opened 2020-12-29
fix(image): svg image src may be a canvas element.

plainheart:
Related source code in ECharts: http://github.com/apache/incubator-echarts/tree/master/src/chart/heatmap/HeatmapLayer.ts

**The broken heatmap layer**

image

**After fix**

image

**The related test case can be found in `echarts/test/heatmap-map.html`**


pissang (approved):

pissang:
I was concerning use `toDataURL` each frame may be too slow. But I guess there is no better way.

PR #717, opened 2021-02-04
chore: add `none` to `textStyle` `overflow`

susiwen8:
related to https://github.com/apache/echarts/pull/14214


pissang:
Seems we don't handle `'none'` currently. `null` is used to disable overfull

susiwen8:
> Seems we don't handle `'none'` currently. `null` is used to disable overfull

@pissang should we remove 'none' in [doc](https://echarts.apache.org/zh/option.html#tooltip.textStyle.overflow)?

pissang:
@susiwen8 yes it makes sense

pissang (approved):

PR #719, opened 2021-02-05
fix: clone object extends null

Saber2pr:
echarts options的source/data有时候会使用高速JSON序列化库来生成，例如json-bigint等，生成的对象没有原型。
下面是一个会发生错误的案例：
```ts
import {
    each, clone, map, isTypedArray, setAsPrimitive, isArray, isObject
    // , HashMap , createHashMap, extend, merge,
} from 'zrender/src/core/util';

const source = []
const obj = Object.create(null)
obj.product = 'Matcha Latte'
obj.count = 82
source.push(obj)

clone(source) // Error: source.hasOwnProperty is not defined.
```

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -29,6 +29,7 @@ const TYPED_ARRAY: {[key: string]: boolean} = {
 };
 
 const objToString = Object.prototype.toString;
+const objHasOwnProperty = Object.prototype.hasOwnProperty
 
 const arrayProto = Array.prototype;
 const nativeForEach = arrayProto.forEach;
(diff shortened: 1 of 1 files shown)


pissang:
These codes are usually being hotspots in common scenarios. But this change seems will slow down the `hasOwnProperty` check due to this benchmark
https://www.measurethat.net/Benchmarks/Show/11476/0/objectprototypehasownproperty-vs-objhasownproperty

I think it's better to a deep clone outside to have a proper prototype on the objects.

plainheart:
Or we check if the `hasOwnProperty` function exists first? https://jsben.ch/bKRaY

pissang:
https://www.measurethat.net/Benchmarks/Show/11481/0/objectprototypehasownproperty-vs-objhasownproperty-vs-e

The performance will also drop if do null check.

image

plainheart:
Yeap, the running result of the benchmark shows obviously any extra judgment can slow the speed down. 😄

PR #720, opened 2021-02-08
fix(env): fix potential NPE in `env.canvasSupported`.

plainheart:
if 
```js
env.domSupported = typeof document !== 'undefined';
```
is `false`, then `document` may be null, this will bring a NPE by 
```js
env.canvasSupported = !!document.createElement('canvas').getContext
```

UPDATE: It seems that `document` will never be undefined here?

### DIFF
--- a/src/core/env.ts
+++ b/src/core/env.ts
@@ -81,7 +81,7 @@ function detect(ua: string, env: Env) {
         browser.weChat = true;
     }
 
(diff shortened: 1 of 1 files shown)


pissang:
It's in the browser branch. So `document` should not be `undefined`.

But I'm not sure when will `document` be `null` but not `undefined`

plainheart:
Just keep it unchanged.

PR #721, opened 2021-02-08
feat(env): add detection for `transformSupported` & `transform3dSupported`.

plainheart:
For apache/echarts#14246

References about check
- https://github.com/Leaflet/Leaflet/blob/master/src/dom/DomUtil.js
- https://github.com/Leaflet/Leaflet/blob/master/src/core/Browser.js


src/core/env.ts
```diff
@@ -87,6 +89,21 @@ function detect(ua: string, env: Env) {
     env.pointerEventsSupported = 'onpointerdown' in window
         && (browser.edge || (browser.ie && +browser.version >= 11));
     env.domSupported = typeof document !== 'undefined';
+
+    const style = document.documentElement.style;
+    env.ieTransform3dSupported = browser.ie && 'transition' in style;
```
pissang, line 94: I'm not sure, but according to the code in https://github.com/apache/echarts/pull/14246/files#diff-323f00b5085e5da04e37591f992fbb0737db751436bec2667c1de6792f82f33dR116

Does `ieTransform3dSupported ` mean only support 2d transform actually? If so, I will suggest using `transformSupported` and `transform3DSupported` to avoid confusing

plainheart: Right. We may need to improve the naming and detection logic.

100pah (approved):

PR #725, opened 2021-02-24
fix(src/Element.ts) Element type error

lzxb:
Detail see: #724

### DIFF
--- a/src/Element.ts
+++ b/src/Element.ts
@@ -676,7 +676,7 @@ class Element<Props extends ElementProps = ElementProps> {
         return '#fff';
     }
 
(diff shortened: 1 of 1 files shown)


pissang:
Thanks! Seems `getInsideTextFill` and `getOutsideFill` also needs this change.

lzxb:
I tested it. No

lzxb:
I use it to develop a real-time chart, which is very good

pissang:
I mean these methods are similar in design. Their returns should all be nullable.

lzxb:
I see you. Subclass inheritance can be reset to `undefined`

PR #728, opened 2021-02-28
fix the animation from NaN(source value) to number(target value)

Nick22nd:
fix apache/echarts#14281

### DIFF
--- a/src/Element.ts
+++ b/src/Element.ts
@@ -1849,6 +1849,9 @@ function animateToShallow<T>(
                 );
             }
             else {
+                if (isNaN(source[innerKey])) {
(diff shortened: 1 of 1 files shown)


pissang:
@Nick22nd Thanks for the fix.

`source[innerKey]` can be an array or color string, which is also NaN. In fact, invalid value in animation is handled in the `Animator` module:

https://github.com/ecomfe/zrender/blob/master/src/animation/Animator.ts#L345
https://github.com/ecomfe/zrender/blob/master/src/animation/Animator.ts#L122

I think we missed the NaN check in the code leads to the problem

https://github.com/ecomfe/zrender/blob/master/src/animation/Animator.ts#L381

PR #759, opened 2021-05-17
fix(pattern): remove non-required properties from PatternObject and fix some type issues.

plainheart:
fixes apache/echarts#14953

1) make `type`, `svgElement`, `svgWidth`, `svgHeight` of `PatternObject ` optional.
2) fixed some type issues.

### DIFF
--- a/src/graphic/Pattern.ts
+++ b/src/graphic/Pattern.ts
@@ -3,10 +3,9 @@ import { ImageLike } from '../core/types';
 type ImagePatternRepeat = 'repeat' | 'repeat-x' | 'repeat-y' | 'no-repeat'
 
 export interface PatternObjectBase {
+    id?: number
     // type is now unused, so make it optional
     type?: 'pattern'
-
-    image: ImageLike | string
 
     x?: number
     y?: number
@@ -16,6 +15,7 @@ export interface PatternObjectBase {
 }
 
 export interface ImagePatternObject extends PatternObjectBase {
+    image: ImageLike | string
     repeat?: ImagePatternRepeat
 }
 
@@ -25,8 +25,6 @@ export interface InnerImagePatternObject extends ImagePatternObject {
 }
 
 export interface SVGPatternObject extends PatternObjectBase {
-    // id is now only used by SVG renderer
-    id?: number
     /**
      * svg element can only be used in svg renderer currently.
      * svgWidth, svgHeight defines width and height used for pattern.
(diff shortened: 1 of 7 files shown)


src/graphic/Pattern.ts
```diff
@@ -5,16 +5,16 @@ type CanvasPatternRepeat = 'repeat' | 'repeat-x' | 'repeat-y' | 'no-repeat'
 export interface PatternObject {
     id?: number
 
-    type: 'pattern'
+    type?: 'pattern'
 
     image: ImageLike | string
```
pissang, line 10: Is it better to use `union of interface`?
```ts
interface CanvasPatternObject {
  type: 'pattern'
  image: ImageLike | string
  repeat: ...
}
interface SVGPattenObject  {
  type: 'pattern'
  svgElement?: SVGElement
  ....
}

type PatternObject = CanvasPatternObject | SVGPatternObject
```

plainheart: I'm not sure which way is better,

1) Use union types (you mentioned above)
2) Define a parent type named `PatternObject`, `CanvasPatternObject ` and `SVGPatternObject` extend it.

pissang: Union types provide a more precise type checking for the developers. It's usually better when the subtypes are enumerable.

For `Gradient` I now prefer the interface to be:

```ts
export interface GradientObjectBase {
    id?: number
    type: string
    colorStops: GradientColorStop[]
    global: boolean
}

export interface LinearGradientObject extends GradientObjectBase { };
export interface RadiaGradientObject extends GradientObjectBase { };
export type GradientObject = LinearGradientObject | RadiaGradientObject;
```

src/svg/helper/PatternManager.ts
```diff
@@ -122,7 +122,7 @@ export default class PatternManager extends Definable {
      * @param pattern zr pattern instance
      * @param patternDom DOM to update
      */
-    updateDom(pattern: PatternObject, patternDom: SVGElement) {
+    updateDom(pattern: SVGPatternObject, patternDom: SVGElement) {
```
pissang, line 125: Sorry for the confusing name `CanvasPatterObject`. The difference between these two pattern object is whether to use an image in the pattern or use an SVG element. Both of them are supported in SVG renderer. https://github.com/ecomfe/zrender/blob/d30bcf75748ffb40cec43794fddcf89c88660258/src/svg/helper/PatternManager.ts#L139

So perhaps using `ImagePatternObject` and `SVGPatternObject` is clearer.

plainheart: Not sure if 3c2fb9a understands what you mean.

pissang: And SVGPatternObject should not have `image` property.

pissang (approved):

PR #763, opened 2021-05-22
[PENDING] fix(svg): use `textBaseline` specified in text style instead of fixed default value.

plainheart:
**THIS IS A PENDING CHANGE**

I noticed the text node's position rendered by the SVG renderer currently differs from the canvas renderer.

Here are some cases,

I'm not sure about the reason for setting the `dominant-baseline` attribute of the text node as a fixed `middle`(previously) or `central`(currently) in SVG renderer, but if we use `textBaseline` specified in text style, it will render the text as the same as the canvas renderer. At least, it will look quite obviously better than now.

### DIFF
--- a/src/svg/graphic.ts
+++ b/src/svg/graphic.ts
@@ -414,13 +414,13 @@ const svgText: SVGProxy<TSpan> = {
         setTransform(textSvgEl, el.transform);
 
         // Consider different font display differently in vertial align, we always
-        // set vertialAlign as 'middle', and use 'y' to locate text vertically.
+        // set verticalAlign as 'middle', and use 'y' to locate text vertically.
         const x = style.x || 0;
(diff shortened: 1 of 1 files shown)


pissang:
@plainheart Sorry about the late reply.  Text is centered by adjusting y in https://github.com/ecomfe/zrender/blob/b3ead0b2ce7755b30b337fae6c983d7d4bcc8042/src/svg/graphic.ts#L419 manually. I can't remember the reason exactly. Seems it's because other `baseline` configuration may have unexpected result in some font-family like `Microsoft Yahei`. @100pah Do you still remember the detail?

100pah:
> I can't remember the reason exactly. Seems it's because other baseline configuration may have unexpected result in some font-family like Microsoft Yahei. 

Yes, I remember for Canvas renderer, it seems not easy to control make the same effect when font family changed if using different value of `textBaseline`.

PR #784, opened 2021-07-12
fix(canvas): fix unexpected `none` or `null` value for ctx.strokeStyle & ctx.fillStyle

plainheart:
To resolve apache/echarts#15238.

Because unexpected value `null` or `'none'` is assigned to `ctx.strokeStyle` or `ctx.fillStyle`, the Firefox devtools will print lots of annoying CSS warnings.

### DIFF
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -20,9 +20,9 @@ import { REDARAW_BIT, SHAPE_CHANGED_BIT } from '../graphic/constants';
 const pathProxyForDraw = new PathProxy(true);
 
 // Not use el#hasStroke because style may be different.
-function styleHasStroke(style: PathStyleProps) {
+function styleHasStroke(style: PathStyleProps, excludeLineWidth?: boolean) {
     const stroke = style.stroke;
-    return !(stroke == null || stroke === 'none' || !(style.lineWidth > 0));
+    return stroke != null && stroke !== 'none' && (excludeLineWidth || style.lineWidth > 0);
 }
 
 function styleHasFill(style: PathStyleProps) {
@@ -459,14 +459,14 @@ function bindPathAndTextCommonStyle(
             flushPathDrawn(ctx, scope);
             styleChanged = true;
         }
-        ctx.fillStyle = style.fill as string;
+        ctx.fillStyle = styleHasFill(style) ? style.fill : '';
     }
     if (forceSetAll || style.stroke !== prevStyle.stroke) {
         if (!styleChanged) {
             flushPathDrawn(ctx, scope);
             styleChanged = true;
         }
-        ctx.strokeStyle = style.stroke as string;
+        ctx.strokeStyle = styleHasStroke(style, true) ? style.stroke : '';
     }
     if (forceSetAll || style.opacity !== prevStyle.opacity) {
         if (!styleChanged) {
--- a/src/canvas/graphic.ts
+++ b/src/canvas/graphic.ts
@@ -20,9 +20,17 @@ import { REDARAW_BIT, SHAPE_CHANGED_BIT } from '../graphic/constants';
 const pathProxyForDraw = new PathProxy(true);
 
 // Not use el#hasStroke because style may be different.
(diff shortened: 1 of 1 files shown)


src/canvas/graphic.ts
```diff
@@ -20,9 +20,9 @@ import { REDARAW_BIT, SHAPE_CHANGED_BIT } from '../graphic/constants';
 const pathProxyForDraw = new PathProxy(true);
 
 // Not use el#hasStroke because style may be different.
-function styleHasStroke(style: PathStyleProps) {
+function styleHasStroke(style: PathStyleProps, excludeLineWidth?: boolean) {
```
pissang, line 23: I think it will have better performance to use another function that not consider `lineWidth`. See "Optimizing Sorting - Argument Adaptation" section in the [Maybe you don't need Rust and WASM to speed up your JS](https://mrale.ph/blog/2018/02/03/maybe-you-dont-need-rust-to-speed-up-your-js.html) article.

pissang (changes requested):

src/canvas/graphic.ts
plainheart: I'm not sure. The reason why I want to exclude line width is that some legend-related cases can't pass. Please run VRTT to see the broken cases.

plainheart: And thanks for sharing the awesome article. It's very helpful.

pissang: > I'm not sure. The reason why I want to exclude line width is that some legend-related cases can't pass. Please run VRTT to see the broken cases.

This is a bit of a surprise. Do you mind giving one fail test name so I don't have to switch from my current branch(which has a lot of breaks) and run all test cases

plainheart: `area-stack`, `area2`, `axis`, `axis-interval2`, `axisPosition` etc.

pissang: Figured out why. It happens when switching from a style `{ stroke: '#ccc', lineWidth: 0}` to a another `{ stroke: '#ccc', lineWidth: 2}`.  Previous style didn't set `strokeStyle` because `lineWidth` is 0. Then the following style also didn't set it because it compares the `stroke` property with the previous and thought the `strokeStyle` doesn't change. 

To simplify the logic, we should ignore the factor of `lineWidth`.

plainheart: I'm glad we found the reason. I just found another potential warning `Expected color but found '['` when color is gradient. So I check if stroke/fill color is valid before setting strokeStyle/fillStyle.

src/canvas/graphic.ts
```diff
@@ -20,9 +20,17 @@ import { REDARAW_BIT, SHAPE_CHANGED_BIT } from '../graphic/constants';
...
-    return stroke != null && stroke !== 'none' && (excludeLineWidth || style.lineWidth > 0);
+    return !(stroke == null || stroke === 'none' || !(style.lineWidth > 0));
+}
+
+// ignore lineWidth and must be string
+// Expected color but found '[' when color is gradient
+function isValidStrokeFillStyle(
+    strokeOrFill: PathStyleProps['stroke'] | PathStyleProps['fill']
+): strokeOrFill is string {
+    return strokeOrFill != null && strokeOrFill !== 'none' && typeof strokeOrFill === 'string';
```
pissang, line 33: I think we can ignore the `string` check. It won't happen only if the user sets a non-string `color`. But in this case, the check can be very complex, for example, the string is not a valid color string.

We can stop at eliminating the warning from ourself code.

src/canvas/graphic.ts
pissang: Saw this explanation right after I commented about why doing the string check😃

pissang (approved):

plainheart:
I just want to say we can remove the null check.
`typeof strokeOrFill === 'string' && strokeOrFill !== 'none';` is enough.

pissang:
@plainheart We can improve it in another PR😅

PR #790, opened 2021-08-01
change: the judge of the cpx and cpy

HCLacids:


pissang:
@HCLacids Can you describe the reason to do this change?

HCLacids:
> @HCLacids Can you describe the reason to do this change?

Because it turns to NaN, not null in the echarts when the self-pointed Node needs four nodes to render the edge. If I change the value of the cpx2 and cpy2, it will make the label out of control.

pissang:
@HCLacids I'm afraid it will also ignore the 0 value.

HCLacids:
> @HCLacids I'm afraid it will also ignore the 0 value.

How about "cpx2 == null || cpy == nul || cpx2 == NaN || cpy == NaN" or "cpx2 == NaN || cpy2 == NaN"

pissang:
Also I think it's a problem in echarts. We should avoid fixing it by changing the design of underlying zrender

HCLacids:
> Also I think it's a problem in echarts. We should avoid fixing it by changing the design of underlying zrender. Especially it's a breaking change

But in the BezierCurve class in zrender, the cpy2 and cpx2 are only the number type, not the null type

Issue #794, opened 2021-08-10
[DISCUSS] Consider using the standard offsetX/offsetY instead of layerX/layerY for modern versions of Firefox


plainheart:
Here is a stale caution in the following code,

https://github.com/ecomfe/zrender/blob/master/src/core/event.ts#L62-L75

According to [CANIUSE](https://caniuse.com/?search=offsetX) and the [release note](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/39) of Firefox, the `offsetX` and `offsetY` have been supported since desktop Firefox 39 

> Support for MouseEvent.offsetX and MouseEvent.offsetY have been added on desktop (bug 69787, but not on Firefox for Android or Firefox OS (they will be added in Firefox 43).

And they have been also supported by Firefox for Android and Firefox OS since [v43](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/43)

> Support for MouseEvent.offsetX and MouseEvent.offsetY have been activated on Firefox for Android and Firefox OS (bug 1204841).

### One consideration why I want to propose to change

I can't dispatch a manual event with expected `offsetX` & `offsetY` through the constructor of `MouseEvent`.

Since zrender is still using the stale un-standard layerX/layerY in Firefox, I can't construct a `MouseEvent` object through

```js
const event = new MouseEvent('click', {
    bubbles: false,
    layerX: 500, // it's private and readonly for Firefox, anyone can't specify it with the constructor
    layerY: 600
})
```

### Motivation

I'm planning to support mixed event responding in the echarts for AMap plugin: [echarts-extension-amap](https://github.com/plainheart/echarts-extension-amap)

All the time we can only choose one interactive layer from **echarts layer** and **map layer** because of the z-index issue.
But it's possible to get both of them interactive through re-dispatch event to another layer when one layer is clicked, scrolled and etc.

So I tried this feature in the branch [feat-mixed-events](https://github.com/plainheart/echarts-extension-amap/tree/feat-mixed-events). It can be working well in Chrome, but can't work in Firefox for the above issue.

<img src="https://user-images.githubusercontent.com/26999792/128802625-4cae896d-9466-44f4-8df5-ecd1d64603c5.gif" width="400">

A [demo](https://all-my-life.cn/tmp/firefoxtest/index2.html) is available here.

PR #812, opened 2021-09-01
fix(svg): svg mouse event doesn't work normally in Firefox when using shadow.

plainheart:
Fix apache/echarts#15646
Fix apache/echarts#15697

It seems the `layerX/layerY` is unexpected in the case where the chart uses SVG renderer and has shadow filter.

From the above picture, we can see that the values of `layerX/layerY`(unexpected) are not the same as `offsetX/offsetY`(expected) and don't look like the offset relative to the "closest positioned element". (I couldn't know the reason)

According to [CANIUSE](https://caniuse.com/?search=offsetX) and the [release note](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/39) of Firefox, the `offsetX` and `offsetY` have been supported since desktop Firefox 39 

> Support for MouseEvent.offsetX and MouseEvent.offsetY have been added on desktop (bug 69787, but not on Firefox for Android or Firefox OS (they will be added in Firefox 43).

And they have been also supported by Firefox for Android and Firefox OS since [v43](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/43)

> Support for MouseEvent.offsetX and MouseEvent.offsetY have been activated on Firefox for Android and Firefox OS (bug 1204841).

As mentioned above, I'd like to use `offsetX/offsetY` for modern Firefox.

A reproducible demo:

```js
// Please use SVG renderer
option = {
  title: {
    text: '某站点用户访问来源',
    subtext: '纯属虚构',
    left: 'center'
  },
  tooltip: {
    trigger: 'item'
  },
  legend: {
    orient: 'vertical',
    left: 'left',
  },
  series: [
    {
      name: '访问来源',
      type: 'pie',
      radius: '50%',
      data: [
        {value: 1048, name: '搜索引擎'},
        {value: 735, name: '直接访问'},
        {value: 580, name: '邮件营销'},
        {value: 484, name: '联盟广告'},
        {value: 300, name: '视频广告'}
      ],
      emphasis: {
        itemStyle: {
          shadowBlur: 10,
          shadowOffsetX: 0,
          shadowColor: 'red'
        }
      }
    }
  ]
};
```

In addition, if we change to use `offsetX/offsetY`, it will make the feature mentiond in #794 possible.

### DIFF
--- a/src/core/event.ts
+++ b/src/core/event.ts
@@ -67,6 +67,8 @@ export function clientToLocal(
     // <https://bugs.jquery.com/ticket/8523#comment:14>
     // BTW3, In ff, offsetX/offsetY is always 0.
     else if (env.browser.firefox
+        // use offsetX/offsetY for Firefox >= 39
+        // PENDING: consider Firefox for Android and Firefox OS? >= 43
         && env.browser.version < '39'
         && (e as FirefoxMouseEvent).layerX != null
         && (e as FirefoxMouseEvent).layerX !== (e as MouseEvent).offsetX
@@ -74,7 +76,7 @@ export function clientToLocal(
         out.zrX = (e as FirefoxMouseEvent).layerX;
         out.zrY = (e as FirefoxMouseEvent).layerY;
     }
-    // For IE6+, chrome, safari, opera. (When will ff support offsetX?)
+    // For IE6+, chrome, safari, opera, firefox >= 39
     else if ((e as MouseEvent).offsetX != null) {
         out.zrX = (e as MouseEvent).offsetX;
         out.zrY = (e as MouseEvent).offsetY;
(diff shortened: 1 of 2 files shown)


src/core/event.ts
```diff
@@ -67,13 +67,16 @@ export function clientToLocal(
     // <https://bugs.jquery.com/ticket/8523#comment:14>
     // BTW3, In ff, offsetX/offsetY is always 0.
     else if (env.browser.firefox
+        // use offsetX/offsetY for Firefox >= 39
+        // PENDING: consider Firefox for Android and Firefox OS? >= 43
+        && env.browser.version < '39'
```
pissang, line 72: I suppose '39' should be a number here?

pissang:
I'm really impressed by how you research the specs and fix these corner cases. Again thanks a lot!

src/core/event.ts
plainheart: I'm not sure if the version string in some earlier versions will probably be like `88.0.1`, though I saw the version string in the `navigator.userAgent` is like `88.0` in most of the versions. (Tested it in Firefox 50, 70, 80, 92)

pissang: Thanks for the explanation. Seems string comparison is better. Let's keep it.

pissang (approved):

PR #822, opened 2021-09-10
fix(svg): svg shadow to solve apache/echarts#12614

Ovilia:
apache/echarts#12614 reports an SVG rendering bug when all line series data have the same value and no line was drawn if has shadow.


pissang:
Merged this PR to https://github.com/ecomfe/zrender/pull/792

PR #826, opened 2021-09-15
fix prototype pollution in merge, clone, extend utilities

pissang:


100pah (approved):

100pah:
memo poc

```js
// poc.js
const echarts = require('echarts');
const obj = {};
const payload =  JSON.parse('{"__proto__":{"polluted":"Polluted!"}}');

console.log("Before: " + obj.polluted);
echarts.util.merge(obj, payload);
console.log("After: " + polluted);
```

PR #842, opened 2021-10-20
fix(bar-race): extra prop for label value animation

Ovilia:
Add `value` property in the `TextProps` for ECharts bar-race animation.


Ovilia:
Close this PR since this specific type change should be defined in Apache ECharts.

PR #853, opened 2021-11-26
fix(svg): update `_svgDom` when vnode changed.

plainheart:
The initial [`_svgDom`](https://github.com/ecomfe/zrender/blob/next/src/svg/Painter.ts#L73) will be removed and replaced by the newly generated element after the [patch](https://github.com/ecomfe/zrender/blob/next/src/svg/patch.ts#L344) process so that all next operations on the old `_svgDom` won't work, e.g. `resize`, `innerHTML`, `clear`.

### DIFF
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -108,6 +108,7 @@ class SVGPainter implements PainterBase {
             vnode.attrs.style = 'user-select:none;position:absolute;left:0;top:0;';
             patch(this._oldVNode || this._svgDom, vnode);
             this._oldVNode = vnode;
+            this._svgDom = vnode.elm as SVGElement;
         }
     }
 
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -9,16 +9,15 @@ import Displayable from '../graphic/Displayable';
 import Storage from '../Storage';
 import { PainterBase } from '../PainterBase';
 import {
+    createElement,
     createVNode,
     vNodeToString,
     SVGVNodeAttrs,
     SVGVNode,
-    createElement,
-    SVGNS,
-    XLINKNS,
     getCssString,
     BrushScope,
-    createBrushScope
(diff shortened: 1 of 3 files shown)


src/svg/Painter.ts
```diff
@@ -108,6 +108,7 @@ class SVGPainter implements PainterBase {
             vnode.attrs.style = 'user-select:none;position:absolute;left:0;top:0;';
             patch(this._oldVNode || this._svgDom, vnode);
             this._oldVNode = vnode;
+            this._svgDom = vnode.elm as SVGElement;
```
pissang, line 111: In which case the svg dom will be changed? 

I'm concerning if it will cause bug if it's changed after top code getting it through `getSvgDom`

plainheart: Made an online demo for this. Please refer to https://codepen.io/plainheart/pen/BawBeMx.

pissang: Seems it's not expected. Take a deep look at, seems it's because if we doing the first patch and using the container as patch target instead of a vnode in
https://github.com/ecomfe/zrender/blob/9fb1577705ac21e15deb4fc8562a5ca907398cb4/src/svg/Painter.ts#L109

It will force create an empty vnode for this container in https://github.com/ecomfe/zrender/blob/9fb1577705ac21e15deb4fc8562a5ca907398cb4/src/svg/patch.ts#L330

Which has a different `key` with the new vnode and cause elm recreated. Perhaps we create the `_oldVNode` ahead when svgDom is created.

pissang:
https://github.com/ecomfe/zrender/blob/9fb1577705ac21e15deb4fc8562a5ca907398cb4/src/svg/patch.ts#L330

This branch can be removed since we don't use `Element` as patch target anymore.

plainheart:
`insertedVnodeQueue` seems unused as well.

pissang (approved):

PR #856, opened 2021-12-10
feat(svg): support exporting base64-encoded dataURL.

plainheart:
This PR enables the SVG renderer to export base64-encoded dataURL instead of only the URL-encoded UTF-8 string.

### DIFF
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -19,7 +19,7 @@ import {
     createBrushScope,
     createSVGVNode
 } from './core';
-import { normalizeColor } from './helper';
+import { normalizeColor, encodeBase64 } from './helper';
 import { defaults, extend, keys, logError, map } from '../core/util';
 import Path from '../graphic/Path';
 import patch, { updateAttrs } from './patch';
@@ -340,11 +340,13 @@ class SVGPainter implements PainterBase {
         this._oldVNode = null;
     }
     toDataURL(base64?: boolean) {
-        const str = encodeURIComponent(this.renderToString());
+        let str = encodeURIComponent(this.renderToString());
         const prefix = 'data:image/svg+xml;';
-        return base64
-            ? prefix + 'base64,' + window.btoa(unescape(str))
-            : prefix + 'charset=UTF-8,' + str;
+        if (base64) {
+            str = encodeBase64(str);
+            return str ? prefix + 'base64,' + str : null;
+        }
+        return prefix + 'charset=UTF-8,' + str;
     }
 
     refreshHover = createMethodNotSupport('refreshHover') as PainterBase['refreshHover'];
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -340,9 +340,13 @@ class SVGPainter implements PainterBase {
         this._oldVNode = null;
     }
     toDataURL(base64?: boolean) {
-        const str = encodeURIComponent(this.renderToString());
-        return 'data:image/svg+xml;'
-            + (base64 ? 'base64,' + encodeBase64(str) : 'charset=UTF-8,' + str);
+        let str = encodeURIComponent(this.renderToString());
+        const prefix = 'data:image/svg+xml;';
+        if (base64) {
+            str = encodeBase64(str);
+            return str ? prefix + 'base64,' + str : null;
+        }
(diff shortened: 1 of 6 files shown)


src/svg/Painter.ts
```diff
@@ -339,10 +339,12 @@ class SVGPainter implements PainterBase {
...
     }
-    toDataURL() {
-        const str = this.renderToString();
-        const html = encodeURIComponent(str);
-        return 'data:image/svg+xml;charset=UTF-8,' + html;
+    toDataURL(base64?: boolean) {
+        const str = encodeURIComponent(this.renderToString());
+        const prefix = 'data:image/svg+xml;';
+        return base64
+            ? prefix + 'base64,' + window.btoa(unescape(str))
```
pissang, line 346: I think we need to check if window exits here if we are using it in the environment like NodeJS or MiniApp

src/svg/Painter.ts
```diff
@@ -340,11 +340,13 @@ class SVGPainter implements PainterBase {
...
     toDataURL(base64?: boolean) {
-        const str = encodeURIComponent(this.renderToString());
+        let str = encodeURIComponent(this.renderToString());
         const prefix = 'data:image/svg+xml;';
-        return base64
-            ? prefix + 'base64,' + window.btoa(unescape(str))
-            : prefix + 'charset=UTF-8,' + str;
+        if (base64) {
+            str = encodeBase64(str);
+            return str ? prefix + 'base64,' + str : null;
```
pissang, line 347: I'm not sure which one is better if base64 encode is not supported. Still return the string before encode or return null.

Seems return null will be better because developers can check the return easily. If they still wan't the string before encode. They can call this method again without base64.

plainheart: > Seems return null will be better because developers can check the return easily

Yes. If it returns the unencoded string with the `base64` option, the developer may be confused.

pissang: Agree, we can give an error log.

src/svg/helper.ts
```diff
@@ -183,6 +183,9 @@ export const encodeBase64 = (function () {
         };
     }
     return function (str: string): string {
+        if (process.env.NODE_ENV !== 'production') {
+            throw new Error('Base64 isn\'t natively supported in the current environment.');
```
pissang, line 187: Use console.error will be better? Or code like
```ts
toDataURL(true) || toDataURL();
```
won't be executed properly in the dev environment

plainheart: Tweaked.

src/svg/helper.ts
```diff
@@ -169,4 +169,23 @@ export function getSRTTransformString(
     }
 
     return res.join(' ');
-}
\ No newline at end of file
+}
+
+export const encodeBase64 = (function () {
+    if (typeof window !== 'undefined' && typeof window.btoa === 'function') {
```
pissang, line 175: Can use `util.isFunction(window.btoa)` here

src/svg/Painter.ts
```diff
@@ -339,10 +339,14 @@ class SVGPainter implements PainterBase {
...
-        const str = this.renderToString();
-        const html = encodeURIComponent(str);
-        return 'data:image/svg+xml;charset=UTF-8,' + html;
+    toDataURL(base64?: boolean) {
+        let str = encodeURIComponent(this.renderToString());
+        const prefix = 'data:image/svg+xml;';
+        if (base64) {
+            str = encodeBase64(str);
+            return str ? prefix + 'base64,' + str : null;
+        }
```
pissang, line 348: Perhaps `return str && prefix + 'base64,' + str` will be neater here?

src/svg/helper.ts
pissang: I see there is code `typeof window !== 'undefined'` in several places. Should we provide a `hasGlobalWindow` variable to do the check in the `env.ts`?

plainheart: Done. And I'm not sure if it's already enough to assign `hasGlobalWindow` to `true` in [this branch](https://github.com/ecomfe/zrender/pull/856/files#diff-6975367242befd32078190cf2e12e55b9ecc20e377632b2671b260ef3033f263L44).

pissang (approved):

PR #861, opened 2021-12-22
feat(ssr): add useViewBox opt

pissang:
Add new `useViewBox` opts in `renderToString` to make generated SVG responsive. See https://github.com/apache/echarts/issues/16127#issuecomment-983514232

### DIFF
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -20,7 +20,7 @@ import {
     createSVGVNode
 } from './core';
 import { normalizeColor, encodeBase64 } from './helper';
-import { defaults, extend, keys, logError, map, retrieve2 } from '../core/util';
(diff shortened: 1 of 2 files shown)


src/svg/Painter.ts
```diff
@@ -193,7 +198,8 @@ class SVGPainter implements PainterBase {
         return vNodeToString(this.renderToVNode({
             animation: defaultOpts.cssAnimation,
             willUpdate: false,
-            compress: true
+            compress: true,
+            useViewBox: opts.useViewBox
```
plainheart, line 202: One question: Can it be better if `useViewBox` default to `true`?

pissang: Make sense. Updated.

plainheart (approved):

PR #862, opened 2021-12-23
Add .js extension in the import statement of es modules.

pissang:
To allow us importing ES modules in the `lib` folder on the browsers.

Example:

```html
<script type="importmap">
{
    "imports": {
        "zrender/": "https://unpkg.com/zrender@latest/",
        "tslib": "https://unpkg.com/tslib@2.3.1/tslib.es6.js"
    }
}
</script>
<script type="module">
if (typeof process === 'undefined' || !process.env) {
    window.process = {
        env: {
            NODE_ENV: 'development'
        }
    };
}
import PathProxy from 'zrender/lib/core/PathProxy.js'
</script>

```


build/processLib.js
```diff
@@ -0,0 +1,63 @@
...
+    else if (moduleName.endsWith('.js')) {
+        return moduleName;
+    }
+    else {
+        return moduleName + '.js'
+    }
+}
+
+async function transform() {
+    const libFiles = await globby(path.join(__dirname, '../lib/**/*.js'));
```
plainheart, line 30: I can remember I mentioned somewhere a compatibility issue of `globby` in Windows.

It's suggested to use the `cwd` option to specify the scan directory, or it can't match any files in Windows.


```js
const libFiles = await globby([
    '**/*.js'
], {
    cwd: path.join(__dirname, '../lib'),
    absolute: true
})
```

pissang: Thanks for the remind! I did some search and found it seems to be related to this issue https://github.com/sindresorhus/globby/issues/179#issuecomment-990398774 . Just fixed as you suggested.

plainheart (approved):

PR #865, opened 2021-12-29
feat(sector): enhance sector corner radius to support configuring any one corner.

plainheart:
The corner radius for the sector has been supported since v5, but it was limited to a pair of corners, either an inner pair or an outer pair. I can remember @Ovilia asked about it at that time.

This PR enhances this feature so that we can configure the radius of four corners at will. As shown below, the sector has four different corner radius.

A related test case is `sector.html`.

---

Moreover, in commit 53819e99ea5f5b4132d6eb1b4bfe23c95f30b8ea, I tweaked the watch script to process lib extension immediately after the source was compiled.


src/graphic/helper/roundSector.ts
```diff
@@ -89,6 +82,46 @@ function computeCornerTangents(
...
+            arr = [cr[0], cr[0], cr[1], cr[1]];
+        }
+        else {
+            arr = [cr[0], cr[0], 0, 0];
+        }
+    }
+    else {
+        arr = [cr, cr, cr, cr];
+    }
+    return map(arr, (cr, idx) => parsePercent(cr, idx < 2 ? r0 : r));
```
pissang, line 122: Should we support percentage value on the sector `cornerRadius`? Seems `Rect` didn't support it.

Fixed `cornerRadius` seems more common in case. I will prefer keep it simpler here

plainheart: The percentage value is supported from the very beginning of this feature([see here](https://echarts.apache.org/option.html#series-pie.itemStyle.borderRadius)), but previously this logic was put into the upper ECharts, so I just keep it here for the backward compatibility.

As for the necessity of supporting percentage value, I would think out the case in which the developer hopes to get a rounded sector(like a capsule or sausage?) with the percentage of '50%'. Just like the rectangle, we used to set the `border-radius` to be `50%` to get a circle. If the percentage value is not supported, the developer will have to calculate the half value via their specified `r/r0`. That doesn't look so convenient.

plainheart: UPDATE: It usually won't reach half of the radius, as we restrict the max value of the corner radius...

pissang: Didn't notice it has been supported. But it's a bit confusing if it use percentage from `r0` and `r1` respectively. From my side it should be `r1 - r0` if we want `'50%'` can be a perfect value to get a round cap on both side. Is it by design that the corner radius of inner side will be different from the outer side if we use a percentage value?

plainheart: Yes. If it's annular, `r1 - r0` is the real relative value, otherwise, it's `r` or `r0`. Perhaps, we should use `r1 - r0` to avoid confusion when using a percentage value.

src/graphic/helper/roundSector.ts
```diff
@@ -138,38 +169,36 @@ export function buildPath(ctx: CanvasRenderingContext2D | PathProxy, shape: {
         arc = mathAbs(tmpAngles[0] - tmpAngles[1]);
     }
 
-    const cx = shape.cx;
-    const cy = shape.cy;
-    const cornerRadius = shape.cornerRadius || 0;
-    const innerCornerRadius = shape.innerCornerRadius || 0;
+    const [icrStart, icrEnd, ocrStart, ocrEnd] = normalizeCornerRadius(cornerRadius, innerRadius, radius);
```
pissang, line 172: Could we optimize it a bit and not do this normalizing if `cornerRadius` is not given?

src/graphic/helper/roundSector.ts
pissang: Agree.

plainheart: Fixed.

src/graphic/helper/roundSector.ts
plainheart: Fixed.

src/graphic/helper/roundSector.ts
```diff
@@ -89,6 +82,48 @@ function computeCornerTangents(
...
+            arr = cr.concat(cr[len - 1]);
+        }
+        else if (len === 2) {
+            arr = [cr[0], cr[0], cr[1], cr[1]];
+        }
+        else {
+            arr = [cr[0], cr[0], 0, 0];
+        }
+    }
+    else {
```
Ovilia, line 119: Is it possible for `cr` to be `undefined`?

src/graphic/shape/Sector.ts
```diff
@@ -9,8 +9,20 @@ export class SectorShape {
     startAngle = 0
     endAngle = Math.PI * 2
     clockwise = true
-    cornerRadius = 0
-    innerCornerRadius = 0
```
Ovilia, line 13: Do we need to compat with `innerCornerRadius` in order versions?

Ovilia (commented):
So glad to see this PR. Two minor comments which I am not very sure of. @pissang Please have a double check.

src/graphic/helper/roundSector.ts
plainheart: There is a check before this normalization.

src/graphic/shape/Sector.ts
plainheart: The API doesn't change in ECharts, I'm not sure if zrender also needs the backward compatibility.

pissang: I think zrender can do this kind of breaking change to keep the simplicity

pissang: I think zrender can do this kind of breaking change to keep the simplicity

src/graphic/helper/roundSector.ts
```diff
@@ -138,38 +174,42 @@ export function buildPath(ctx: CanvasRenderingContext2D | PathProxy, shape: {
...
+        const { x, y } = calcCircleCenter(cx, cy, radius, startAngle);
+        ctx.moveTo(x, y);
         ctx.arc(cx, cy, radius, startAngle, endAngle, !clockwise);
 
         if (innerRadius > e) {
-            ctx.moveTo(
-                cx + innerRadius * mathCos(endAngle),
-                cy + innerRadius * mathSin(endAngle)
-            );
+            const { x, y } = calcCircleCenter(cx, cy, innerRadius, endAngle);
```
pissang, line 188: I will prefer the original code here. From my side it's pretty intuitive already. Using function is worse on the performance because there is an extra call and will create a temporary object. Also the compiled code will be longer:
```js
// Compiled result.
var a=b(c,d,e,f),x=a.x,y=a.y;ctx.moveTo(x,y);
// Previous
ctx.moveTo(a+b*c(d),a+b*c(d));
```

plainheart: Good catch! It has been recoverd.

src/graphic/helper/roundSector.ts
```diff
@@ -165,11 +189,24 @@ export function buildPath(ctx: CanvasRenderingContext2D | PathProxy, shape: {
     }
     // is a circular or annular sector
     else {
+        let icrStart;
```
pissang, line 192: Should we give them a default `0` vaue? Seems undefined initial value will lead to some `NaN` values, e.g. `ocrs`, `ocre`, `icrs`, `icre` seems will be `NaN`. I'm not sure if it will cause unexpected issues.

plainheart: I think it's unnecessary. I made this change in the commit 3bb9b1be6758ecda9503fc786927dd5077004684. All the conditions about numeric comparison won't pass as `NaN > e` is false.

https://github.com/ecomfe/zrender/blob/3bb9b1be6758ecda9503fc786927dd5077004684/src/graphic/helper/roundSector.ts#L225

https://github.com/ecomfe/zrender/blob/3bb9b1be6758ecda9503fc786927dd5077004684/src/graphic/helper/roundSector.ts#L254

https://github.com/ecomfe/zrender/blob/3bb9b1be6758ecda9503fc786927dd5077004684/src/graphic/helper/roundSector.ts#L288

pissang (approved):

PR #880, opened 2022-01-19
fix(sector): fix a case in which the angle is NaN.

plainheart:
In #870, we fixed the animation bug caused by the `NaN` value. But that only works in canvas renderer. In SVG renderer, the path can't be successfully built because of the `NaN` value.

The related test case is the same as #870, that is, the last test case in echarts/test/pie-animation.html

### DIFF
--- a/src/graphic/helper/roundSector.ts
+++ b/src/graphic/helper/roundSector.ts
@@ -145,11 +145,13 @@ export function buildPath(ctx: CanvasRenderingContext2D | PathProxy, shape: {
         innerRadius = tmp;
     }
 
+    const { startAngle, endAngle } = shape;
+    if (isNaN(startAngle) || isNaN(endAngle)) {
+        return;
+    }
+
+    const { cx, cy } = shape;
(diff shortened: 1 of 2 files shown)


plainheart:
BTW, I found another bug related to label & labelLine. It should be related to these three places.

1. https://github.com/apache/echarts/blob/release/src/chart/pie/labelLayout.ts#L395-L399
2. https://github.com/apache/echarts/blob/release/src/chart/pie/labelLayout.ts#L552-L555
3. https://github.com/apache/echarts/blob/release/src/chart/pie/labelLayout.ts#L564-L567

It didn't take into account the case in which the data becomes normal from NaN value.

Here is an online test case: https://codepen.io/plainheart/pen/ZEXNQVy

pissang:
I think shape with NaN value should be not displayed. If we treat it same with 0. Border will still show.

plainheart:
@pissang You're right. Fixed.

pissang (approved):

PR #891, opened 2022-02-21
fix(dirtyRect):  text dirty rect considered when text in host

linghaoSu:
- Resolve https://github.com/apache/echarts/issues/16534

### Why
when enabling dirty rect, and then clicking the button in the toolbox, the zrender consider where should to repainted, but does not consider text when they are hosted in another element, so, maybe we should consider ignore textElment when they hosted in other element and the host element that is ignored.

### Before 
image

### After
image

### DIFF
--- a/src/graphic/Displayable.ts
+++ b/src/graphic/Displayable.ts
@@ -195,7 +195,8 @@ class Displayable<Props extends DisplayableProps = DisplayableProps> extends Ele
         viewWidth: number,
         viewHeight: number,
         considerClipPath: boolean,
-        considerAncestors: boolean
+        considerAncestors: boolean,
+        considerHostTarget: boolean
     ) {
         const m = this.transform;
         if (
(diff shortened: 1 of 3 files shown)


src/graphic/Displayable.ts
```diff
@@ -234,6 +235,23 @@ class Displayable<Props extends DisplayableProps = DisplayableProps> extends Ele
             }
         }
 
+        // consider host target
+        if (
```
pissang, line 239: We can treat `__hostTarget` as a special `parent` and doing both check in the

https://github.com/ecomfe/zrender/blob/458d111917e520ac944f6541169990db164f6ccb/src/graphic/Displayable.ts#L229
```ts
if (considerAncestors) {
      let node = this;
      while (node) {
          if (node.ignore) {
              return false;
          }
          node = node.parent || node.__hostTarget;
      }
  }
```

linghaoSu: Wow, I agree with this way, we don't need to modify the api of `shouldBePainted`, I have fixed this using these codes. Please review again.

linghaoSu:
hi, @pissang Is there anything else wrong with this pr

PR #898, opened 2022-03-09
fix: Uncaught TypeError: Failed to execute 'createRadialGradient' on …

lefex:
…'CanvasRenderingContext2D': The provided double value is non-finite.

    at createRadialGradient. Excute npm run test error fix! (#16649)

The bug in Echarts https://github.com/apache/echarts/issues/16649

### DIFF
--- a/src/canvas/helper.ts
+++ b/src/canvas/helper.ts
@@ -4,7 +4,7 @@ import { GradientObject } from '../graphic/Gradient';
 import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
-export function safeNum(num: number, defalutNum: number = 0, negative = false) {
+export function safeNum(num: number, defalutNum: number = 0, nonNegative = false) {
     // null
     if (num === null) {
         return defalutNum;
@@ -14,7 +14,7 @@ export function safeNum(num: number, defalutNum: number = 0, negative = false) {
         return defalutNum;
     }
     // non-negative
-    if (negative && num < 0) {
+    if (nonNegative && num < 0) {
         return defalutNum;
     }
 
--- a/src/canvas/helper.ts
+++ b/src/canvas/helper.ts
@@ -5,10 +5,11 @@ import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
 export function safeNum(num: number, defalutNum: number = 0, negative = false) {
-    // null NaN or empty string
-    if (num === null || isNaN(num) || `${num}` === '') {
+    // null
+    if (num === null) {
         return defalutNum;
     }
+    // NaN, Infinity, undefined, 'xx'
     if (!isFinite(num)) {
         return defalutNum;
     }
@@ -16,7 +17,8 @@ export function safeNum(num: number, defalutNum: number = 0, negative = false) {
     if (negative && num < 0) {
         return defalutNum;
     }
-    return +num;
+
+    return num;
 }
 
 export function createLinearGradient(
--- a/src/canvas/helper.ts
+++ b/src/canvas/helper.ts
@@ -9,6 +9,9 @@ export function safeNum(num: number, defalutNum: number = 0, negative = false) {
     if (num === null || isNaN(num) || `${num}` === '') {
         return defalutNum;
     }
+    if (!isFinite(num)) {
+        return defalutNum;
+    }
     // non-negative
     if (negative && num < 0) {
         return defalutNum;
--- a/src/canvas/helper.ts
+++ b/src/canvas/helper.ts
@@ -4,21 +4,21 @@ import { GradientObject } from '../graphic/Gradient';
 import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
-export function safeNum(num: number, defalutNum: number = 0, nonNegative = false) {
+export function isSafeNum(num: number, nonNegative = false) {
     // null
     if (num === null) {
-        return defalutNum;
+        return false;
     }
     // NaN, Infinity, undefined, 'xx'
     if (!isFinite(num)) {
-        return defalutNum;
+        return false;
     }
     // non-negative
     if (nonNegative && num < 0) {
-        return defalutNum;
+        return false;
     }
 
-    return num;
+    return true;
 }
 
 export function createLinearGradient(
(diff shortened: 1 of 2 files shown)


src/canvas/helper.ts
```diff
@@ -4,30 +4,36 @@ import { GradientObject } from '../graphic/Gradient';
 import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
+export function safeNum(num: number, defalutNum: number = 0, negative = false) {
+    // null NaN or empty string
+    if (num === null || isNaN(num) || `${num}` === '') {
```
pissang: Seems `isNaN(num)` also includes `null` and `''` check.

src/canvas/helper.ts
```diff
@@ -4,30 +4,41 @@ import { GradientObject } from '../graphic/Gradient';
 import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
+export function safeNum(num: number, defalutNum: number = 0, negative = false) {
```
pissang, line 7: Use name `nonNegative` will be better.

plainheart:
If we filter these illegal values in the underlying zrender without any warnings, will it hide potential errors of arguments passed from the upper application like ECharts? Most of the time, I think such an issue is caused by error arguments from the callers.

src/canvas/helper.ts
lefex: have changed

src/canvas/helper.ts
lefex: have changed

lefex:
> If we filter these illegal values in the underlying zrender without any warnings, will it hide potential errors of arguments passed from the upper application like ECharts? Most of the time, I think such an issue is caused by error arguments from the callers.

I think we should check the parameters to make sure there are no errors.

plainheart:
It may be more difficult for the developers to find out why they didn't get the expected result as no errors were thrown and no warnings were output.

pissang:
We can have warning logs rather exception when having invalid values.

Ovilia:
How about giving warning as well as doing such safe guarding?
You can use `warn` function in `util/log` to do so.

lefex:
> How about giving warning as well as doing such safe guarding? You can use `warn` function in `util/log` to do so.

ok！

lefex:
I add a waring: [PR](https://github.com/ecomfe/zrender/pull/898/files)  I'm not going to change the `createLinearGradient` method [.](https://github.com/ecomfe/zrender/blob/master/src/canvas/helper.ts#L25).

src/canvas/helper.ts
```diff
@@ -4,6 +4,23 @@ import { GradientObject } from '../graphic/Gradient';
 import { RectLike } from '../core/BoundingRect';
 import Path from '../graphic/Path';
 
+export function isSafeNum(num: number, nonNegative = false) {
+    // null
+    if (num === null) {
+        return false;
+    }
+    // NaN, Infinity, undefined, 'xx'
+    if (!isFinite(num)) {
```
pissang, line 13: `!isFinite` also includes `num === null` check.

These three check can be merged to:

```ts
if (!isFinite(num) || (nonNegative && num < 0 {
  ...
}
```

src/canvas/helper.ts
```diff
@@ -46,12 +63,18 @@ export function createRadialGradient(
...
+
     if (!obj.global) {
         x = x * width + rect.x;
         y = y * height + rect.y;
         r = r * min;
     }
 
+    if (!isSafeNum(x) || !isSafeNum(y) || !isSafeNum(r, true)) {
+        console.warn('The provided value is non-finite. You must provide x and y.');
+        return;
```
pissang, line 75: return nothing here is not safe. Because we expect this method to return a CanvasGradient and set it to CanvasRenderingContext. If we return `null`, it will be an undefined behavior.

I think we can fallback to default value when the `x`, `y`, `r` is invalid like `createLinearGradient` did.

Also there are several concerns about this warn message being to frequently:
1. It should be warn only once if the `CanvasGradientObject` is not changed. Use `WeakMap` to mark this `CanvasGradientObject` has been warned may be a solution.
2. The `boundingRect` sometimes may be infinite or NaN(when we don't want to show the element) and leading this `x`, `y`, `r` to be invalid. In this case, the error message will not be helpful but confuse the 
developers. So it's better to check the `obj.x`, `obj.y`, `obj.r` earlier and give warning message. Then we do check `x`, `y`, `r` again and fallback to default value.
3. The check of `obj.x`, `obj.y`, `obj.r` and warning message can be done only when `process.env.NODE_ENV !== 'production'` so these can be ommitted in the production environment.

At last the `createLinearGradient` should do the same.

src/canvas/helper.ts
lefex: image

pissang: Sorry it's my mistake. `null` can be convert to `0` here. I rechecked if `null` can be used as argument of `createLinearGradient` and seems it can. But `undefined` will throw an `non-finite` exception. So if we mainly wan using this check to avoid exceptions, we can still remove `null` check.

src/canvas/helper.ts
lefex: The obj is changed when mouse hover, can not use WeakMap to add warning log only once. So I added a guard to ensure that the parameters are valid.

image

src/canvas/helper.ts
```diff
@@ -46,12 +51,17 @@ export function createRadialGradient(
...
+
     if (!obj.global) {
         x = x * width + rect.x;
         y = y * height + rect.y;
         r = r * min;
     }
 
+    x = isSafeNum(x) ? x : 0.5;
+    y = isSafeNum(y) ? y : 0.5;
+    r = isSafeNum(r, true) ? r : 0.5;
```
pissang, line 63: Seems only one number needs to be check positive. In this case I will prefer keeping the `isSafeNum` simpler and more cohesion that only accept one parameter `number`. Monomorphic function is always better than polymorphic function, on both code style and performance.

So the `r` check can be
```ts
r = isSafeNum(r) && r >= 0 ? r : 0.5
```

test/ut/spec/contain/Sector.test.ts
```diff
@@ -12,7 +12,7 @@ describe('Path', function () {
                 cy: 625.5,
                 startAngle: -1.5707963267948966,
                 endAngle: 4.71238898038469,
-                innerCornerRadius: 5,
```
pissang, line 15: This change seems to be unnecessary?

pissang (approved):

PR #918, opened 2022-05-25
Feature: coarse pointer tolerance

Ovilia:
### Intent

This PR is intended to increase elements' responsive area on mobile devices so that elements can be touched more easily.

The idea is inspired by W3C's [Understanding Success Criterion 2.5.5: Target Size](https://www.w3.org/WAI/WCAG21/Understanding/target-size.html).

### Algorithm

The algorithm can be briefly described as:

1. Find the hovering element at the pointer position. If it exists, return the hovering element. Otherwise, go to 2.
2. Calculate the elements whose bounding boxes have intersection with the touch position enlarged by "pointerSize" (which is 44px by default)
3. Loop through different radius and angle from the pointer position to check if there are elements intersecting with it.

For cases that have too many elements, this feature can be turned off by setting `ignoreCoarsePointer` to the elements.

Tested with the Apache ECharts test cases and in most cases the extra calculation time is under 0.1ms.

### API

```ts
zrender.init({
  useCoarsePointer: 'auto',
  pointerSize: 44,
  ...
});
```

- `useCoarsePointer` (`'auto'` | `true` | `false`): whether to enable thie feature. If set to be `'auto'`, it's turned on only for mobile devices.
- `pointerSize` (`number`): pointer position is increased by this size in diameter.

Each element can be set with `ignoreCoarsePointer` to turn off this feature individually:

```ts
el.ignoreCoarsePointer = true;
```

### DIFF
--- a/src/Handler.ts
+++ b/src/Handler.ts
@@ -132,6 +132,8 @@ const handlerNames = [
 type HandlerName = 'click' |'dblclick' |'mousewheel' |'mouseout' |
     'mouseup' |'mousedown' |'mousemove' |'contextmenu';
 
+const tmpRect = new BoundingRect(0, 0, 0, 0);
+
 // TODO draggable
 class Handler extends Eventful {
 
@@ -368,14 +370,17 @@ class Handler extends Eventful {
             const pointerSize = this._pointerSize;
             const targetSizeHalf = pointerSize / 2;
             const pointerRect = new BoundingRect(x - targetSizeHalf, y - targetSizeHalf, pointerSize, pointerSize);
+
             for (let i = list.length - 1; i >= 0; i--) {
                 if (list[i] !== exclude
                     && !list[i].ignore
                     && !list[i].ignoreTargetSize
                 ) {
-                    const rect = list[i].getBoundingRect().clone();
-                    rect.applyTransform(list[i].transform);
-                    if (rect.intersect(pointerRect)) {
+                    tmpRect.copy(list[i].getBoundingRect());
+                    if (list[i].transform) {
+                        tmpRect.applyTransform(list[i].transform);
+                    }
+                    if (tmpRect.intersect(pointerRect)) {
                         candidates.push(list[i]);
                     }
                 }
--- a/src/Handler.ts
+++ b/src/Handler.ts
@@ -354,26 +354,51 @@ class Handler extends Eventful {
             }
         }
 
+        // console.time('targetSize');
         if (!out.target) {
-            // If no element at pointer position, increase targetSize
+            /**
+             * If no element at pointer position, check intersection with
+             * elements with pointer enlarged by target size.
+             */
+            const candidates = [];
             const targetSize = 44;
-            let minSize = Number.MAX_VALUE;
+            const targetSizeHalf = targetSize / 2;
+            const pointerRect = new BoundingRect(x - targetSizeHalf, y - targetSizeHalf, targetSize, targetSize);
             for (let i = list.length - 1; i >= 0; i--) {
-                const rect = list[i].getBoundingRect();
-                const size = Math.min(rect.width, rect.height);
-                if (
-                    list[i] !== exclude
+                if (list[i] !== exclude
                     && !list[i].ignore
                     && !list[i].silent
-                    && size < targetSize
-                    && size < minSize
-                    && list[i].targetSizeContain(x, y, targetSize)
                 ) {
-                    out.target = list[i];
-                    minSize = size;
+                    const rect = list[i].getBoundingRect().clone();
+                    rect.applyTransform(list[i].transform);
+                    if (rect.intersect(pointerRect)) {
+                        candidates.push(list[i]);
+                    }
+                }
+            }
+
+            if (candidates.length) {
+                const rStep = 4;
+                const thetaStep = Math.PI / 12;
+                for (let r = 0; r < targetSizeHalf && !out.target; r += rStep) {
+                    for (let theta = 0; theta < Math.PI * 2 && !out.target; theta += thetaStep) {
+                        const x1 = x + r * Math.cos(theta);
+                        const y1 = y + r * Math.sin(theta);
+                        for (let i = 0; i < candidates.length; i++) {
+                            const hoverCheckResult = isHover(candidates[i], x1, y1);
+                            if (hoverCheckResult) {
+                                !out.topTarget && (out.topTarget = list[i]);
+                                if (hoverCheckResult !== SILENT) {
+                                    out.target = list[i];
+                                    break;
+                                }
+                            }
+                        }
+                    }
                 }
             }
         }
+        // console.timeEnd('targetSize');
 
         return out;
     }
--- a/src/Handler.ts
+++ b/src/Handler.ts
@@ -147,6 +147,8 @@ class Handler extends Eventful {
 
     private _draggingMgr: Draggable
 
+    private _targetSize: number
+
     _downEl: Element
     _upEl: Element
     _downPoint: [number, number]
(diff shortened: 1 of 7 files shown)


src/Handler.ts
```diff
@@ -353,6 +361,61 @@ class Handler extends Eventful {
...
+            /**
+             * If no element at pointer position, check intersection with
+             * elements with pointer enlarged by target size.
+             */
+            const candidates = [];
+            const pointerSize = this._pointerSize;
+            const targetSizeHalf = pointerSize / 2;
+            const pointerRect = new BoundingRect(x - targetSizeHalf, y - targetSizeHalf, pointerSize, pointerSize);
+
+            for (let i = list.length - 1; i >= 0; i--) {
```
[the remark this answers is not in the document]
Ovilia: Thanks for reminding!

src/core/event.ts
```diff
@@ -309,6 +309,9 @@ export function isMiddleOrRightButtonOnMouseUpDown(e: { which: number }) {
     return e.which === 2 || e.which === 3;
 }
 
+export function isTouchDevice() {
+    return ('ontouchstart' in window) || (navigator.maxTouchPoints > 0);
```
pissang, line 313: `window` and `navigator` should be checked if exists here.

src/zrender.ts
```diff
@@ -126,7 +127,16 @@ class ZRender {
         const handerProxy = (!env.node && !env.worker && !ssrMode)
             ? new HandlerProxy(painter.getViewportRoot(), painter.root)
             : null;
-        this.handler = new Handler(storage, painter, handerProxy, painter.root);
+
+        const useTargetSize = (opts.useCoarsePointer == null || opts.useCoarsePointer === 'auto')
```
pissang, line 131: `opts.useCoarsePointer` can be cached to a variable

src/zrender.ts
```diff
@@ -126,7 +127,16 @@ class ZRender {
...
+            ? isTouchDevice()
+            : !!opts.useCoarsePointer;
+        const defaultTargetSize = 44;
+        const targetSize = useTargetSize
```
pissang, line 135: `targetSize` and `useTargetSize` should be renamed to `pointerSize` and `usePointerSize`?

src/Handler.ts
```diff
@@ -339,20 +347,77 @@ class Handler extends Eventful {
...
+
+            /**
+             * If there are elements whose bounding boxes are near the pointer,
+             * use the most top one that intersects with the enlarged pointer.
+             */
+            if (candidates.length) {
+                const rStep = 4;
+                const thetaStep = Math.PI / 12;
+                for (let r = 0; r < targetSizeHalf && !out.target; r += rStep) {
+                    for (let theta = 0; theta < Math.PI * 2 && !out.target; theta += thetaStep) {
```
pissang, line 403: Because in the inner loop it will break after assigning to `out.target`. There is no need to check `out.target` here

src/Handler.ts
```diff
@@ -339,20 +347,77 @@ class Handler extends Eventful {
...
+                        const x1 = x + r * Math.cos(theta);
+                        const y1 = y + r * Math.sin(theta);
+                        for (let i = 0; i < candidates.length; i++) {
```
pissang, line 406: This logic is same to https://github.com/ecomfe/zrender/blob/85d6614900e127c3abcc81fef58f496e14fe7eba/src/Handler.ts#L375 . Perhaps we can extract it to a separate function to avoid writing this complex `target`, `topTarget` and `SILENT` things twice.

And don't forget to reverse the iteration order in https://github.com/ecomfe/zrender/pull/918/files#diff-2befe9e583e17eed633cc9bd738c9b0c54ffb1f48222871eda2cc4dc232c48eeR375 after this change

pissang:
The whole design looks good. There are several optimizations we can do to improve the code. I commented about them.

src/Handler.ts
Ovilia: I think this is still necessary because `break` can only break one level of inner loop. Can you double check if I understand it right?

plainheart: Why not directly return the `out` target when it's found? We can move the judgment from two `for` loops to the back of `setHoverTarget` function.

```js
for (outer loop) {
  for (inner loop) {
    // ....
    setHoverTarget(xxxx);
    if (out.target) {
      return out;
    }
  }
}
```

Ovilia: Good idea

src/Handler.ts
```diff
@@ -337,18 +345,54 @@ class Handler extends Eventful {
...
-                    break;
+            /**
+             * If there are elements whose bounding boxes are near the pointer,
+             * use the most top one that intersects with the enlarged pointer.
+             */
+            if (candidates.length) {
+                const rStep = 4;
+                const thetaStep = Math.PI / 12;
+                for (let r = 0; r < targetSizeHalf; r += rStep) {
+                    for (let theta = 0; theta < Math.PI * 2; theta += thetaStep) {
```
plainheart, line 388: How about caching `Math.PI * 2` to a constant `PI2`?

src/Handler.ts
```diff
@@ -468,6 +512,30 @@ function isHover(displayable: Displayable, x: number, y: number) {
...
+function setHoverTarget(
+    list: Displayable[],
+    out: HoveredResult,
+    x: number,
+    y: number,
+    exclude: Displayable
+) {
+    for (let i = list.length - 1; i >= 0; i--) {
+        const el = list[i];
+        let hoverCheckResult;
```
plainheart, line 524: I think the variable `hoverCheckResult`can be declared outside the `for` loop.

src/zrender.ts
```diff
@@ -126,7 +127,17 @@ class ZRender {
...
-        this.handler = new Handler(storage, painter, handerProxy, painter.root);
+
+        const useCoarsePointer = opts.useCoarsePointer;
+        const usePointerSize = (useCoarsePointer == null || useCoarsePointer === 'auto')
+            ? isTouchDevice()
+            : !!useCoarsePointer;
+        const defaultPointerSize = 44;
+        const targetSize = usePointerSize
+            ? (opts.pointerSize == null ? defaultPointerSize : opts.pointerSize)
+            : 0;
```
plainheart, line 138: `0` looks not good semantically when not using the coarse pointer. I prefer `undefined/null`. 

And this block can be simplified as
```ts
let pointerSize;
if (usePointerSize) {
  pointerSize = retrive2(opts.pointerSize, defaultPointerSize);
}
// or
usePointerSize && (pointerSize = retrive2(opts.pointerSize, defaultPointerSize);
```

Optionally, `defaultPointerSize` can also be replaced with its value `44` and add a comment for it.

src/Handler.ts
```diff
@@ -337,18 +345,54 @@ class Handler extends Eventful {
     findHover(x: number, y: number, exclude?: Displayable): HoveredResult {
         const list = this.storage.getDisplayList();
         const out = new HoveredResult(x, y);
+        setHoverTarget(list, out, x, y, exclude);
```
plainheart, line 348: It seems we can check if the list is empty before this line.

src/core/event.ts
```diff
@@ -309,6 +309,10 @@ export function isMiddleOrRightButtonOnMouseUpDown(e: { which: number }) {
     return e.which === 2 || e.which === 3;
 }
 
+export function isTouchDevice() {
+    return (typeof window === 'object' && 'ontouchstart' in window)
```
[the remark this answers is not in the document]
Ovilia: OK. I think we can use `touchEventsSupported` here. If its implementation should be improved, we can update it in ZRender later.

pissang:
I have no more questions on this PR

plainheart (approved):
LGTM

PR #928, opened 2022-06-15
chore: avoid fork repository run github ci

btea:


plainheart (approved):
LGTM. Thanks for your contribution! We may need to do the same work for ECharts later.

/cc @pissang

pissang:
Thanks!  

Just created a PR in the ECharts. https://github.com/apache/echarts/pull/17245

Issue #965, opened 2022-10-19
HashMap.each performance


JonasBa:
Hello, we (@getsentry) have recently started profiling one of our services which runs echarts to generate thumbnail images of charts so that they can be easily shared as previews and while checking some profiles generated in prod, we noticed that a function called `each` is often at the top of the stack.

<img width="1281" alt="CleanShot 2022-10-19 at 16 22 24@2x" src="https://user-images.githubusercontent.com/9317857/196797516-3543ef70-2673-4a44-a8b5-df5e64a8df6b.png">
I've manually looked up the dist to verify that this is hashmap related.

Looking at [HashMap.each](https://github.com/ecomfe/zrender/blob/master/src/core/util.ts#L694), there could be two possible performance culprits in there.

First is v8 disabling optimization of `for...in` described [here](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#5-for-in), this could occur in the event that our {} dict exceeds the number of keys and the underlying v8 data structure backing the object starts using an actual hash map which makes enumeration slow.

Second, I noticed that map.get does a call to `hasOwnProperty` before returning the value, this seems much slower than just returning the value of `dict[key]`, see [micro benchmark](https://jsbench.me/oel9d7kxt5/1).

I could see this being improved by using native maps as the underlying class and relying on Map.entries for iteration (there is a performance benefit here in the sense that we can also avoid a second function call to `get`) as well as possibly removing the  hasOwnProperty call.

I would like to get your input on this issue, more specifically, I'd like to better understand the motivation behind not using native maps (assuming this could have been due to bad browsers support at the time?). Full disclaimer that I am not very familiar with the rest of the echarts codebase, so I lack context if this is a very commonly used method or if it just relates to the specific charts we generate. I would also be more than happy to contribute myself if we can agree on a direction.

Appreciate any thoughts and it goes without saying, but thank you for maintaining the library that we use daily 👏🏼

plainheart:
Thanks for your investigation and suggestion! It sounds great and we agree on this! :)

> I'd like to better understand the motivation behind not using native maps (assuming this could have been due to bad browsers support at the time?)

You are right. The custom `HashMap` is being used for a long time for the compatibility of IE 10 and lower. As far, I think we can switch to the native `Map`, and meanwhile, make a polyfill for the legacy IE as we still hope the library can support IE 10 ~ 11.

JonasBa:
[...]

PR #966, opened 2022-10-20
feat(polyfill): use native map if available

JonasBa:
Related to the discussion in https://github.com/ecomfe/zrender/issues/965

This PR is lacking tests, if the general direction is ok and validated, I will add the relevant tests.

From initial benchmarks
```
> node src/core/util.benchmark.js
cpu: Apple M1 Pro
runtime: node v18.3.0 (arm64-darwin)

benchmark      time (avg)             (min … max)       p75       p99      p995
------------------------------------------------- -----------------------------
• n=8
------------------------------------------------- -----------------------------
Native      92.59 ns/iter  (88.37 ns … 521.94 ns)  94.09 ns  116.1 ns 131.65 ns
Polyfill   140.06 ns/iter (137.22 ns … 171.45 ns) 139.03 ns 164.76 ns 170.73 ns

summary for n=8
  Native
   1.51x faster than Polyfill

• n=32
------------------------------------------------- -----------------------------
Native      298.8 ns/iter    (292 ns … 649.44 ns) 294.09 ns 630.26 ns 649.44 ns
Polyfill     1.17 µs/iter     (1.15 µs … 1.19 µs)   1.17 µs   1.19 µs   1.19 µs

summary for n=32
  Native
   3.9x faster than Polyfill

• n=64
------------------------------------------------- -----------------------------
Native     558.55 ns/iter (554.65 ns … 581.92 ns)  558.3 ns 580.21 ns 581.92 ns
Polyfill      2.4 µs/iter     (2.38 µs … 2.43 µs)   2.41 µs   2.43 µs   2.43 µs

summary for n=64
  Native
   4.3x faster than Polyfill

• n=128
------------------------------------------------- -----------------------------
Native       1.09 µs/iter     (1.08 µs … 1.13 µs)   1.09 µs   1.13 µs   1.13 µs
Polyfill      4.8 µs/iter     (4.76 µs … 4.84 µs)   4.81 µs   4.84 µs   4.84 µs

summary for n=128
  Native
   4.41x faster than Polyfill

• n=1024
------------------------------------------------- -----------------------------
Native       8.47 µs/iter     (8.46 µs … 8.57 µs)   8.47 µs   8.57 µs   8.57 µs
Polyfill    43.66 µs/iter  (42.83 µs … 236.92 µs)  43.54 µs  46.92 µs  48.67 µs

summary for n=1024
  Native
   5.15x faster than Polyfill

• n=8192
------------------------------------------------- -----------------------------
Native      68.23 µs/iter    (67.5 µs … 93.42 µs)  68.25 µs  72.46 µs  73.46 µs
Polyfill   771.05 µs/iter   (752.29 µs … 1.06 ms) 772.42 µs 871.79 µs 897.46 µs

summary for n=8192
  Native
   11.3x faster than Polyfill

• n=200000
------------------------------------------------- -----------------------------
Native       1.66 ms/iter     (1.64 ms … 1.71 ms)   1.66 ms    1.7 ms    1.7 ms
Polyfill    28.46 ms/iter    (28.15 ms … 29.6 ms)  28.51 ms   29.6 ms   29.6 ms

summary for n=200000
  Native
   17.17x faster than Polyfill
   ```

### DIFF
--- a/src/core/util.ts
+++ b/src/core/util.ts
@@ -653,20 +653,131 @@ export function isPrimitive(obj: any): boolean {
     return obj[primitiveKey];
 }
 
+interface Iterator<IteratorValue extends unknown, IteratorReturn> {
+    next(): {value: IteratorValue, done: false} | {value: IteratorReturn, done: true};
+}
+
+interface MapInterface<T, KEY extends string | number = string | number> {
+    // Size should be a getter here, but tsconfig will not allow getters (es3 target).
+    // Since our polyfill and HashMap class do not expose this, it should be ok.
+    size(): number;
+    clear(): undefined;
+    delete(key: KEY): boolean;
+    get(key: KEY): T | undefined
+    set(key: KEY, value: T): this
+    has(key: KEY): boolean
+    keys(): Iterator<KEY, undefined>
+    values(): Iterator<T, undefined>
+    entries(): Iterator<[KEY, T], undefined>
+    forEach(callback: (key: KEY, value: T) => void): void
+}
+
+class MapPolyfill<T, KEY extends string | number = string | number> implements MapInterface<T, KEY> {
+    private data: Record<KEY, T> = {} as Record<KEY, T>;
+
+    size(): number {
+        return Object.keys(this.data).length;
+    }
+    toString(): string {
+        return '[object MapPolyfill]';
+    }
+    clear(): undefined {
+        this.data = {} as Record<KEY, T>;
+        return undefined;
+    }
+    delete(key: KEY): boolean {
+        const existed = this.has(key);
+        delete this.data[key];
+        return existed;
+    }
+    get(key: KEY): T | undefined {
+        return this.data[key];
+    }
+    set(key: KEY, value: T): this {
+        this.data[key] = value;
+        return this;
+    }
+    has(key: KEY): boolean {
+        return this.data.hasOwnProperty(key);
+    }
+    keys(): Iterator<KEY, undefined> {
+        const keys = Object.keys(this.data) as KEY[];
+        let i = 0;
+        let end = keys.length;
+        return {
+            next() {
+                if (i < end) {
+                    return {value: keys[i] as KEY, done: false};
+                }
+                return {done: true, value: undefined};
+            }
+        };
+    }
+    values(): Iterator<T, undefined> {
+        const data = this.data;
+        const keys = Object.keys(this.data) as KEY[];
+        let i = 0;
+        let end = keys.length;
+        return {
+            next() {
+                if (i < end) {
+                    return {value: data[keys[i]] as T, done: false};
+                }
+                return {done: true, value: undefined};
+            }
+        };
+    }
+    entries(): Iterator<[KEY, T], undefined> {
+        const data = this.data;
+        const keys = Object.keys(this.data) as KEY[];
+        let i = 0;
+        let end = keys.length;
+        return {
+            next() {
+                if (i < end) {
+                    return {value: [keys[i], data[keys[i]]] as [KEY, T], done: false};
+                }
+                return {done: true, value: undefined};
+            }
+        };
+    }
+    forEach(callback: (key: KEY, value: T) => void): void {
+        // This is a potential performance bottleneck, see details in
+        // https://github.com/ecomfe/zrender/issues/965, however it is now
+        // less likely to occur as we default to native maps when possible.
+        for (const key in this.data) {
+            if (this.data.hasOwnProperty(key)) {
+                callback(key, this.data[key]);
+            }
+        }
+    }
+
+    [Symbol.iterator](): Iterator<[KEY, T], undefined> {
+        return this.entries();
+    }
+}
+
+// We want to use native Map if it is available, but we do not want to polyfill the global scope
+// in case users ship their own polyfills or patch the native map object in any way.
+const isNativeMapSupported = typeof Map === 'function';
+function maybeNativeMap<T, KEY extends string | number = string | number>(): MapInterface<T, KEY> {
+    // Map may be a native class if we are running in an ES6 compatible environment.
+    // eslint-disable-next-line
+    return (isNativeMapSupported ? Map : MapPolyfill) as unknown as MapInterface<T, KEY>;
+}
 
 /**
  * @constructor
- * @param {Object} obj Only apply `ownProperty`.
+ * @param {Object} obj
  */
 export class HashMap<T, KEY extends string | number = string | number> {
-
-    data: {[key in KEY]: T} = {} as {[key in KEY]: T};
+    data: MapInterface<T, KEY> = maybeNativeMap<T, KEY>();
 
     constructor(obj?: HashMap<T, KEY> | { [key in KEY]?: T } | KEY[]) {
         const isArr = isArray(obj);
         // Key should not be set on this, otherwise
         // methods get/set/... may be overrided.
-        this.data = {} as {[key in KEY]: T};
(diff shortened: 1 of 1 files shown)

### CODE REVIEW

File: src/core/util.ts [hunk 1]
```diff
@@ -653,20 +653,131 @@ export function isPrimitive(obj: any): boolean {
...
 
 /**
  * @constructor
- * @param {Object} obj Only apply `ownProperty`.
+ * @param {Object} obj
  */
 export class HashMap<T, KEY extends string | number = string | number> {
-
-    data: {[key in KEY]: T} = {} as {[key in KEY]: T};
+    data: MapInterface<T, KEY> = maybeNativeMap<T, KEY>();
```
plainheart, line 774: `= maybeNativeMap<T, KEY>()` can be removed as it will be assigned in the constructor.

src/core/util.ts
```diff
@@ -682,31 +793,40 @@ export class HashMap<T, KEY extends string | number = string | number> {
...
     // from ES6 Map).
     get(key: KEY): T {
-        return this.data.hasOwnProperty(key) ? this.data[key] : null;
+        return this.data.get(key);
     }
     set(key: KEY, value: T): T {
         // Comparing with invocation chaining, `return value` is more commonly
         // used in this case: `const someVal = map.set('a', genVal());`
-        return (this.data[key] = value);
+        // eslint-disable-next-line
```
plainheart, line 801: It seems this comment can be removed.

src/core/util.ts
```diff
@@ -682,31 +793,40 @@ export class HashMap<T, KEY extends string | number = string | number> {
...
-        }
+        this.data.forEach((key, value) => {
+            cb.call(context, value, key);
+        });
     }
     keys(): KEY[] {
-        return keys(this.data);
+        const keys: KEY[] = [];
+        // We are iterating using the keys iterator and not directly
+        // accesing this.data[key] in our for..in loop. This produces
```
plainheart, line 818: `accesing` -> `accessing`

src/core/util.ts
```diff
@@ -653,20 +653,131 @@ export function isPrimitive(obj: any): boolean {
...
+        return this;
+    }
+    has(key: KEY): boolean {
+        return this.data.hasOwnProperty(key);
+    }
+    keys(): Iterator<KEY, undefined> {
+        const keys = Object.keys(this.data) as KEY[];
+        let i = 0;
+        let end = keys.length;
+        return {
```
pissang, line 707: I think this iterator polyfill can't used in the `for...in..` loop in the browsers don't support iterator because we don't use babel to transpile the code.

We can directly return an array here. And in the `HashMap#keys`, we check if the `data` is an native map and wrap it with `Array.from` if so.

It's a bit dirty but I think it's ok if we don't expose this polyfill and keep it isolated.

src/core/util.ts
```diff
@@ -653,20 +653,70 @@ export function isPrimitive(obj: any): boolean {
...
+    }
+    get(key: KEY): T | undefined {
+        return this.data[key];
+    }
+    set(key: KEY, value: T): this {
+        this.data[key] = value;
+        return this;
+    }
+    keys(): KEY[] {
+        return Object.keys(this.data) as KEY[];
```
plainheart, line 685: Use the `keys` function in the same file for consistency and remove `as KEY[]`.

src/core/util.ts
```diff
@@ -682,31 +732,36 @@ export class HashMap<T, KEY extends string | number = string | number> {
...
     each<Context>(
         cb: (this: Context, value?: T, key?: KEY) => void,
         context?: Context
     ) {
-        for (let key in this.data) {
-            if (this.data.hasOwnProperty(key)) {
-                cb.call(context, this.data[key], key);
-            }
-        }
+        this.data.forEach((key, value) => {
```
plainheart, line 749: For native `Map`, the params of the `forEach` callback function are `value, key`. We can change the callback params of `MapPolyfill#forEach` to `callback: (value: T, key: KEY) => void`.

src/core/util.ts
```diff
@@ -653,20 +653,70 @@ export function isPrimitive(obj: any): boolean {
...
-
-    data: {[key in KEY]: T} = {} as {[key in KEY]: T};
+    data: MapInterface<T, KEY>
 
     constructor(obj?: HashMap<T, KEY> | { [key in KEY]?: T } | KEY[]) {
         const isArr = isArray(obj);
         // Key should not be set on this, otherwise
         // methods get/set/... may be overrided.
-        this.data = {} as {[key in KEY]: T};
+        this.data = maybeNativeMap<T, KEY>();
```
plainheart, line 719: The `maybeNativeMap()` seems to return a class instead of an instance.

src/core/util.ts
```diff
@@ -682,31 +732,36 @@ export class HashMap<T, KEY extends string | number = string | number> {
...
-            }
-        }
+        this.data.forEach((key, value) => {
+            cb.call(context, value, key);
+        });
     }
     keys(): KEY[] {
-        return keys(this.data);
+        // Polyfilled map return and Array<KEYS> for keys
+        if (this.data instanceof MapPolyfill) {
```
plainheart, line 755: Maybe we can use the  `isNativeMapSupported` constant instead of the `instanceof` operator to detect if it's the map polyfill. (Not sure)

src/core/util.ts
JonasBa: Oh yeah, good find

src/core/util.ts
JonasBa: I think we can, but it exposes a possible problem if the user or some 3rd party package is applying polyfills after this module is evaluated. I would tend to say that while possible, it is probably very low risk. We should be able to assume that any polyfills should still correctly implement the native Map interface.

src/core/util.ts
JonasBa: Nice find.

src/core/util.ts
```diff
@@ -653,20 +653,67 @@ export function isPrimitive(obj: any): boolean {
...
+        return this;
+    }
+    keys(): KEY[] {
+        return keys(this.data);
+    }
+    forEach(callback: (value: T, key: KEY) => void): void {
+        // This is a potential performance bottleneck, see details in
+        // https://github.com/ecomfe/zrender/issues/965, however it is now
+        // less likely to occur as we default to native maps when possible.
+        for (const key in this.data) {
```
plainheart, line 688: Here we can cache `this.data` to a constant.

```ts
const data = this.data;
for (const key in data) {
    if (data.hasOwnProperty(key)) {
        callback(data[key], key);
    }
}
```

plainheart (commented):
The others LGTM

src/core/util.ts
JonasBa: I would not expect this to have a noticeable performance impact on performance as we are just referencing to a prototype chain, but I ran a quick micro benchmark to confirm (https://jsbench.me/3vl9wtiswy/1). I would prefer to go for simple code here, but if you have a strong preference please let me know and I will change it.

JonasBa: Oh I see. If you have a strong preference then I wouldn't mind changing it, but I don't think it makes much difference in bundle size

plainheart: It can only reduce several bytes after minification insignificant but we can do it.

plainheart (approved):
Thank you @JonasBa for the nice work! 
@pissang Any other thoughts? If none, we can merge it.

JonasBa:
Appreciate the reviews and helpful comments @plainheart @pissang 🙏

pissang (approved):

PR #968, opened 2022-10-24
fix(platformApi): fix wrong expression for extracting the font size

### LINKED ISSUE #947: measureText 计算 fontSize 有误
https://github.com/ecomfe/zrender/blob/e5d275f5d08b3093e3e218843c593f000f86d62f/src/core/platform.ts#L75-L79

DEFAULT_FONT 为 `12px sans-serif`

运行时为 `normal normal 12px sans-serif`

```js
font = font || DEFAULT_FONT;
const res = /^([0-9]*?)px$/.exec(font);
const fontSize = +(res && res[1]) || DEFAULT_FONT_SIZE;
```

res 为 null, fontSize 只能为 DEFAULT_FONT_SIZE

plainheart:
Fix #947
Fix apache/echarts#17326

The font string might be composed as `normal normal 12px sans-serif` via the [`Text#makeFont`](https://github.com/ecomfe/zrender/blob/master/src/graphic/Text.ts#L894) function. So the `^` and `$` should be removed from the expression.

| Before | After |
| :----: | :----: |
| image | image |

### DIFF
--- a/src/core/platform.ts
+++ b/src/core/platform.ts
@@ -75,8 +75,8 @@ export const platformApi: Platform = {
                 text = text || '';
                 font = font || DEFAULT_FONT;
                 // Use font size if there is no other method can be used.
-                const res = /^([0-9]*?)px$/.exec(font);
-                const fontSize = +(res && res[1]) || DEFAULT_FONT_SIZE;
+                const res = /(\d+)px/.exec(font);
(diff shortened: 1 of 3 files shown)


pissang (approved):

zhiqingchen:
hi @plainheart 
Is there a plan to publish the changes?

PR #994, opened 2023-03-10
chore: release 5.4.3

Ovilia:
When releasing ZRender 5.4.2, I forgot to run `npm run release` before `npm publish` so that the version number in `'lib'` is not correct. So we need to release a new version 5.4.3.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1,12 +1,12 @@
 {
   "name": "zrender",
-  "version": "5.4.2",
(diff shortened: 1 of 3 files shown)


plainheart (approved):

plainheart (changes requested):
The version in src/zrender also
 needs to be updated.

plainheart (approved):

PR #998, opened 2023-03-11
chore: fix deprecated npm script & add check for version and unexpected files before publishing to npm

plainheart:
### Background
According to recent mistakes in publishing to npm, sometimes some files neither tracked by git nor ignored by the `.npmignore` file are unexpectedly published [1], and more often, the versions in different files are inconsistent [2][3], which is more serious than the former. Therefore, I've added a script `build/prepublish.js` that runs before really publishing distribution files to npm to reduce the possibility of this kind of mistake.

> [1] [`1.log`](https://cdn.jsdelivr.net/npm/echarts@5.4.0/1.log) was published to npm in Apache ECharts v5.4.0
> [2] Published v5.4.2 but [the version in `lib/zrender.js`](https://cdn.jsdelivr.net/npm/zrender@5.4.2/lib/zrender.js) is still `5.4.1`
> [3] Published v5.3.1 but both [the version in `lib/zrender.js`](https://cdn.jsdelivr.net/npm/zrender@5.4.2/lib/zrender.js) and [the version in `src/zrender.ts`](https://github.com/ecomfe/zrender/blob/5.3.1/src/zrender.ts#L475) are still `5.3.0`

### Changes

**1. Replace the deprecated `prepublish` hook in favor of `prepare` & `prepublishOnly`.**

**2.  Check the consistency of versions in different key distribution files.**

>    \- **_package.json_**
>    \- **_package-lock.json_**
>    \- **_dist/zrender.js_**
>    \- **_dist/zrender.min.js_**
>    \- **_lib/zrender.js_**
>    \- **_src/zrender.ts_**

This will output the version in each file and the file path to help you confirm if they are all expected.
If the version in any file doesn't match the others, an error will be shown, and it will give the corresponding steps to fix it.

**3. Check if there are unexpected files that probably shouldn't be published.**

   This will list all untracked files by git and check if they have been ignored by the `.npmignore` file.
   If not, an error will be shown and output all unexpected files for double-check.

image

### DIFF
--- a/build/prepublish.js
+++ b/build/prepublish.js
@@ -1,0 +1,88 @@
+const fs = require('fs-extra');
+const chalk = require('chalk');
+const ignore = require('ignore');
+const { execSync } = require('node:child_process');
+
+console.log();
+console.log(chalk.yellowBright(`⚠️ You should have run ${chalk.bold('`npm run release`')} before running this script!`));
+console.log();
+
+// check versions in key dist files
+
+console.log(chalk.yellow('🔎 Checking versions in dist files...'));
+
+const fileVersions = [
+    'package.json',
+    'package-lock.json',
+    'dist/zrender.js',
+    'dist/zrender.min.js'
+].map(filePath => ({
+    file: filePath,
+    version: require('../' + filePath).version
+}));
+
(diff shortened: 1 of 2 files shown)


Ovilia (approved):
This offers great help! Thanks!

PR #999, opened 2023-03-14
feat(ssr): add emphasis style in ssr css apache/echarts#18334

Ovilia:
See apache/echarts#18334

### DIFF
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -128,6 +128,7 @@ class SVGPainter implements PainterBase {
         willUpdate?: boolean
         compress?: boolean,
         useViewBox?: boolean
+        emphasis?: boolean
     }) {
 
         opts = opts || {};
@@ -140,6 +141,7 @@ class SVGPainter implements PainterBase {
         scope.animation = opts.animation;
         scope.willUpdate = opts.willUpdate;
         scope.compress = opts.compress;
+        scope.emphasis = opts.emphasis;
 
         const children: SVGVNode[] = [];
 
@@ -173,7 +175,12 @@ class SVGPainter implements PainterBase {
          * If add css animation.
          * @default true
          */
-        cssAnimation?: boolean
+        cssAnimation?: boolean,
+        /**
+         * If add css emphasis.
+         * @default true
+         */
+        cssEmphasis?: boolean,
         /**
          * If use viewBox
          * @default true
@@ -183,6 +190,7 @@ class SVGPainter implements PainterBase {
         opts = opts || {};
         return vNodeToString(this.renderToVNode({
             animation: retrieve2(opts.cssAnimation, true),
+            emphasis: retrieve2(opts.cssEmphasis, true),
             willUpdate: false,
             compress: true,
             useViewBox: retrieve2(opts.useViewBox, true)
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -124,10 +124,10 @@ class SVGPainter implements PainterBase {
     }
 
     renderToVNode(opts?: {
-        animation?: boolean
-        willUpdate?: boolean
+        animation?: boolean,
+        willUpdate?: boolean,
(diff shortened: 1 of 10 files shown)


src/svg/cssEmphasis.ts
```diff
@@ -0,0 +1,82 @@
...
+
+export function createCSSEmphasis(
+    el: Displayable,
+    attrs: SVGVNodeAttrs,
+    scope: BrushScope
+) {
+    if (!el.ignore && el.__metaData) {
+        const empahsisStyle = el.states.emphasis
+            ? el.states.emphasis.style
+            : {};
```
plainheart, line 41: It seems the [`emphasis.disabled`](https://echarts.apache.org/zh/option.html#series-bar.emphasis.disabled) option is not being respected. The chart still highlights the item that the emphasis state is disabled.

src/svg/cssEmphasis.ts
```diff
@@ -0,0 +1,82 @@
...
+        }
+        let lineWidth = empahsisStyle.lineWidth;
+        if (lineWidth) {
+            // Symbols use transform to set size, so lineWidth
+            // should be divided by scaleX
+            const scaleX = el.transform ? el.transform[0] : 1;
+            lineWidth = lineWidth / scaleX;
+        }
+        const style = {
+            cursor: 'pointer', // TODO: Should be included in el
```
plainheart, line 62: Will this TODO be finished in this PR? Should it be `el.cursor || 'pointer'` here?

src/svg/graphic.ts
```diff
@@ -69,6 +70,14 @@ function setStyleAttrs(attrs: SVGVNodeAttrs, style: AllStyleOption, el: Path | T
     setShadow(el, attrs, scope);
 }
 
+function setMetaData(attrs: SVGVNodeAttrs, el: Path | TSpan | ZRImage) {
+    if (el.__metaData) {
+        for (const key in el.__metaData) {
+            attrs[META_DATA_PREFIX + key] = el.__metaData[key] + '';
+        }
```
plainheart, line 77: The eslint plugin point out that the for-in has no guard. Perhaps, we can use the `each` util?

src/svg/graphic.ts
```diff
@@ -69,6 +70,14 @@ function setStyleAttrs(attrs: SVGVNodeAttrs, style: AllStyleOption, el: Path | T
     setShadow(el, attrs, scope);
 }
 
+function setMetaData(attrs: SVGVNodeAttrs, el: Path | TSpan | ZRImage) {
+    if (el.__metaData) {
+        for (const key in el.__metaData) {
+            attrs[META_DATA_PREFIX + key] = el.__metaData[key] + '';
```
100pah, line 76: Perhaps we should convert the attribute key to lower case explicitly.
See https://stackoverflow.com/a/25033330/6075217

100pah (changes requested):

src/svg/cssEmphasis.ts
Ovilia: I'm not sure if there is a need to customize the cursor when hovering. `el.cursor` already exists (`Displable.cursor`). Maybe `el.hoverCursor`? I'm not too sure about this.

src/svg/Painter.ts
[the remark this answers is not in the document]
Ovilia: `opts.emphasis` is used in [src/svg/graphic.ts](https://github.com/ecomfe/zrender/pull/999/files#diff-d36c603bc219342aa51e6e0381dfac549a0f907ea7fe6777385edfc127f166a3R219)

src/svg/cssEmphasis.ts
Ovilia: I would like to leave this to be like this for this PR and see in the future if it needs to be custermized.

100pah (approved):
LGTM

100pah (approved):

PR #1002, opened 2023-04-03
fix(svg): duplicate id for background rect with multiple charts

Ovilia:
Fix apache/echarts#18194 where the background rect of svg charts are all having the same `id="0"`.

### DIFF
--- a/src/svg/Painter.ts
+++ b/src/svg/Painter.ts
@@ -143,7 +143,7 @@ class SVGPainter implements PainterBase {
 
         const children: SVGVNode[] = [];
 
-        const bgVNode = this._bgVNode = createBackgroundVNode(this._id, width, height, this._backgroundColor, scope);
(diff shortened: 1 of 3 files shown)


src/svg/Painter.ts
```diff
@@ -143,7 +143,7 @@ class SVGPainter implements PainterBase {
 
         const children: SVGVNode[] = [];
 
-        const bgVNode = this._bgVNode = createBackgroundVNode(width, height, this._backgroundColor, scope);
+        const bgVNode = this._bgVNode = createBackgroundVNode(this._id, width, height, this._backgroundColor, scope);
```
plainheart, line 146: I always feel this id is unnecessary as I can't see its useful points and the user can't set it in some way. Perhaps we can remove it directly.

Ovilia:
@plainheart I agree with you. Please review again. Thanks

plainheart (approved):

PR #1006, opened 2023-04-12
Made non-BMP characters be treated as a single character for wrapping.

Wetterquarz:
Characters use two UTF-16 blocks to be represented. They can be taken apart by the wrapping algorithm. This contribution aims to fix that.


Wetterquarz:
For anyone interested in the standard: https://www.unicode.org/reports/tr14/

Issue #1033, opened 2023-09-23
[Feature] Support rotation relative to a point other than the origin


ChepteaCatalin:
Currently, the `rotate` function from `core/matrix.ts` rotates the input matrix by the specified angle relative to the origin point (0, 0) of the coordinate system.

When parsing an SVG that contains a `rotate` transformation (`rotate(<a> [<x> <y>])`), only the angle `a` is taken into consideration, while the coordinates `x` and `y`, which specify the point about which to rotate, are ignored. This fact caused [echarts/19141](https://github.com/apache/echarts/issues/19141) issue.

The expected behavior is that the point about which to rotate is taken into consideration, enabling a rotation relative to a specific point.

PR #1034, opened 2023-09-23
feat(matrix): allow rotating relative to a point different than the origin

ChepteaCatalin:
This PR fixes the issue #1033 and specifically [echarts/19141](https://github.com/apache/echarts/issues/19141).

It adds the possibility to provide pivot coordinates around which to rotate the matrix.

If documentation updates are required, could you please assist me with it? The documents are in Chinese, and I don't understand the language.


Ovilia (approved):
Thanks for your contribution! I will help update the document.

PR #1036, opened 2023-11-02
fix(env): fix env detection compatibility in node v21.1.0

Uzlopak:
This PR improves the detection of nodejs version 21.1.0 and above

### DIFF
--- a/src/core/env.ts
+++ b/src/core/env.ts
@@ -37,7 +37,10 @@ else if (typeof document === 'undefined' && typeof self !== 'undefined') {
     // In worker
     env.worker = true;
 }
-else if (typeof navigator === 'undefined') {
+else if (
+    typeof navigator === 'undefined' ||
+    navigator.userAgent.indexOf('Node.js') === 0
+) {
     // In node
     env.node = true;
     env.svgSupported = true;


Uzlopak:
I personally would have preferred

`if (typeof process !== 'undefined' && process.versions && typeof node.versions.node === 'string')`

plainheart (approved):
Thanks for your contribution!

Related information:
- Release note: https://github.com/nodejs/node/releases/tag/v21.1.0
- PR: https://github.com/nodejs/node/pull/50200

plainheart:
@Uzlopak A small question: Before v21.1.0, is the `navigator.userAgent` undefined? Will it bring an NPE? Do we need to add the existence check?

Uzlopak:
For what stands the abbreviation NPE?

plainheart:
Null Pointer Exception - that is, a type error thrown when accessing properties / methods of an undefined object.

plainheart:
I looked through the [release note of Node v21.0.0](https://github.com/nodejs/node/releases/tag/v21.0.0), which added the `navigator.hardwareConcurrency` property. So in that version, `navigator.userAgent` may be undefined.

Uzlopak:
In node 21.0.0 userAgent was missing. But node 21 is anyway a dev version and it was expected to result in regressions. So using it for production is anyway bad.

I would personally try to detect node/deno/bun with

`if ((typeof process !== 'undefined' && typeof process.version === 'string') || 'Deno' in window)`

first one would detect node and bun, second one would detect Deno.

tniessen:
> But node 21 is anyway a dev version and it was expected to result in regressions. So using it for production is anyway bad.

FWIW, that is _not_ the official stance of the Node.js project. While we generally recommend LTS versions for production, non-LTS release lines are not "dev versions". They typically receive the same level of support as LTS versions, albeit for a shorter amount of time. Breaking changes may be introduced in both LTS and non-LTS release lines, but regressions are highly undesirable in any case.

PR #1053, opened 2024-01-23
fix: fix lint problems and add ci workflow; invTransform NPE due to #1011

Ovilia:


src/core/timsort.ts
```diff
@@ -217,23 +215,12 @@ function gallopRight<T>(value: T, array: T[], start: number, length: number, hin
...
 
-    length = array.length;
-
-    if (length < 2 * DEFAULT_TMP_STORAGE_LENGTH) {
-        tmpStorageLength = length >>> 1;
-    }
-
     var tmp: T[] = [];
 
-    stackLength = length < 120 ? 5 : length < 1542 ? 10 : length < 119151 ? 19 : 40;
```
Ovilia, line 235: Unused parameter. I checked the [original code](https://github.com/mziccard/node-timsort/blob/master/src/timsort.js#L438) that `stackLength` was used to init array but we don't use it so it can be safely removed.

plainheart (approved):

PR #1064, opened 2024-02-07
fix(image): fix image dont trigger onload event when hit cache

xdbobname:
make sense in echarts markpoint symbol, or other symbol


Ovilia (changes requested):
Please add a test case in zrender.

xdbobname:
> Please add a test case in zrender.

Sorry for MIA, I have added one.

PR #1077, opened 2024-05-13
fix(platform): fix `measureText` can't recognize decimal font size

Unlasting1:
I am using echarts in a react native environment.
Which uses rem adaptation, so a similar code appears
fontSize: rem(12) => fontSize: 11.5
I found a problem.When fontSzie is set to decimal it causes some rendering exceptions.
After troubleshooting I found out that it is an issue with a regularity in measureText in zrender. This regular expression was causing abnormal behavior when parsing decimals.


plainheart (approved):
Thanks for your contribution!

PR #1081, opened 2024-06-25
Update BoundingRect intersection check

JesseSDevaney:
- For full context, see the corresponding PR in the ECharts repository: https://github.com/apache/echarts/pull/19982

---
**Bug Fix:** This PR is intended to fix the `hideOverlap` functionality in ECharts by preventing empty labels (`""`) from being considered as candidates for the overlap checking.
- An empty label is also a bounding rect with width and height <= 0


Ovilia:
Closed due to https://github.com/apache/echarts/pull/19982#pullrequestreview-2147585769

PR #1092, opened 2024-08-02
fix(svg): fix SVG rendering bug when gradient color stops includes 'transparent' color

plainheart:
| Before | After |
| :----: | :----: |
| image | image |


Ovilia (dismissed):
LGTM

PR #1101, opened 2024-11-18
feat: support output the state `isTruncated` in `Text` element.

100pah:
support output the state `isTruncated` in `Text` element, which can be used in the upper program like echarts tooltip, as required by apache/echarts#16315 .

Test case: 
test/text-overflow.html


plainheart (approved):
Great job! ✨

PR #1111, opened 2025-01-06
fix(env): improve env detection compatibility for Node

liamjones:
The React Native test environment *is* Node but sets up a window global: https://github.com/facebook/react-native/blob/7a85b911254dbae9fac4c6d8a862a20c33e0b36a/packages/react-native/Libraries/Core/setUpGlobals.js#L18

Reintroduce the navigator.userAgent check as a fallback for detecting a Node environment

### DIFF
--- a/src/core/env.ts
+++ b/src/core/env.ts
@@ -40,7 +40,8 @@ else if (typeof document === 'undefined' && typeof self !== 'undefined') {
 else if (
     !env.hasGlobalWindow
     || 'Deno' in window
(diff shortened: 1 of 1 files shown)


liamjones:
Without this tests in an RN environment fall over on this line:
https://github.com/ecomfe/zrender/blob/ae8cfaae186e6c1bf66b5dc431b2cdda5e67dacf/src/core/env.ts#L90

As there's a `window` but no `document`

plainheart (approved):
Thanks for your contribution!

PR #1116, opened 2025-03-20
fix(svg): fix `stop-opacity` is not parsed by SVG parser

plainheart:
Fix (partially) https://github.com/apache/echarts/issues/18407#issuecomment-2738836773

| Before | After |
| :----: | :----: |
| image | image |

### DIFF
--- a/src/tool/parseSVG.ts
+++ b/src/tool/parseSVG.ts
@@ -19,6 +19,7 @@ import RadialGradient, { RadialGradientObject } from '../graphic/RadialGradient'
 import Gradient, { GradientObject } from '../graphic/Gradient';
 import TSpan, { TSpanStyleProps } from '../graphic/TSpan';
 import { parseXML } from './parseXML';
+import * as colorTool from './color';
 
 
 interface SVGParserOption {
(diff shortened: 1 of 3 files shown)


src/tool/parseSVG.ts
```diff
@@ -627,9 +628,19 @@ function parseGradientColorStops(xmlNode: SVGElement, gradient: GradientObject):
             // <stop stop-color="red"/>
             const styleVals = {} as Dictionary<string>;
             parseInlineStyle(stop, styleVals, styleVals);
-            const stopColor = styleVals.stopColor
+            let stopColor = styleVals.stopColor
                 || stop.getAttribute('stop-color')
                 || '#000000';
+            const stopOpacity = styleVals.stopOpacity
+                || stop.getAttribute('stop-opacity');
+            if (stopOpacity) {
```
pissang, line 636: Should use `stopOpacity != null` to avoid `0` being ignored here.

src/tool/parseSVG.ts
```diff
@@ -627,9 +628,19 @@ function parseGradientColorStops(xmlNode: SVGElement, gradient: GradientObject):
...
+                const rgba = colorTool.parse(stopColor);
+                const stopColorOpacity = rgba && rgba[3];
+                if (stopColorOpacity) {
```
pissang, line 639: Same here

pissang (changes requested):

src/tool/parseSVG.ts
plainheart: Both getAttribute and styleVars are always string type and the string '0' is truthy.

pissang (approved):

PR #1125, opened 2025-06-16
feat: may be an approach to fix #1105

100pah:
Is it works for #1105 ?
But I haven't test compondPath case. Just an idea.
And may need @pissang to review it.


Ovilia (approved):
Tested with custom stage series and confirmed that this can fix #1105. Thanks! :)

PR #1149, opened 2026-01-22
feat(canvas): update DPR after resizing

Justin-ZS:
#### Summary
- Allow dynamic `devicePixelRatio` updates through the `resize()` method
- Add `Layer.updateDpr()` to keep canvas context dpr in sync

#### Fix
https://github.com/apache/echarts/issues/21386

#### Usage

```typescript
zr.resize({
    width: 800,
    height: 600,
    devicePixelRatio: window.devicePixelRatio
});
```

#### Changes
- Add optional `devicePixelRatio` parameter to `PainterBase.resize()` interface
- Add `Layer.updateDpr()` method
- Update `Painter.resize()` to trigger layer resize on dpr change
- Expose `devicePixelRatio` option in `ZRender.resize()`


src/canvas/Painter.ts
```diff
@@ -838,8 +838,14 @@ export default class CanvasPainter implements PainterBase {
      */
     resize(
         width?: number | string,
-        height?: number | string
+        height?: number | string,
+        devicePixelRatio?: number
     ) {
+        const dprChanged = devicePixelRatio != null && devicePixelRatio !== this.dpr;
```
plainheart: Follow up on https://github.com/apache/echarts/pull/21489#pullrequestreview-3726258813, we can apply the same logic to automatically retrieve the current DPR when `resizeOptions.devicePixelRatio` is not specified. Perhaps we should refactor the code at https://github.com/ecomfe/zrender/blob/b013719b035f9b034806fdf160f07849b290c3a8/src/config.ts#L3-L22 into a function?

plainheart: I also noticed the minimum DPR is currently limited to 1 - it can result in a blurry chart when the browser is scaled down. Should we remove this limit?

Justin-ZS: done.
the device dpr is obtained in Painter.resize now, and the minimum dpr has been removed.

https://github.com/Justin-ZS/zrender/blob/5a52347e793a2eeffcc9eaed5dd4538087755eac/src/canvas/Painter.ts#L847-L849

src/canvas/Painter.ts
plainheart: I tend to completely replace the constant `devicePixelRatio` in favor of the new `getDevicePixelRatio()` function.

src/canvas/Layer.ts
plainheart: The same here.

plainheart: Or just let the DPR fall back to 1 and only do it in zrender constructor? 
https://github.com/Justin-ZS/zrender/blob/5a52347e793a2eeffcc9eaed5dd4538087755eac/src/zrender.ts#L92

src/canvas/Painter.ts
Justin-ZS: done

src/canvas/Layer.ts
Justin-ZS: I prefer to do it only in the Canvas Painter constructor.

src/canvas/Painter.ts
```diff
@@ -175,7 +175,6 @@ export default class CanvasPainter implements PainterBase {
                 // TODO sting?
                 height = opts.height as number;
             }
-            this.dpr = opts.devicePixelRatio || 1;
```
Justin-ZS, line 178: @plainheart 
Since `this.dpr` has already been initialized before, I removed this line.
Please let me know if forcing 1 here was intentional.

Justin-ZS:
@plainheart Could you help to review again? Thanks!

Jancheng-z:
这是来自QQ邮箱的假期自动回复邮件。
 
plainheart:
> @plainheart Could you help to review again? Thanks!

Generally, I have no more comments. I plan to merge your PR if there is no review from other members after the release of ECharts v6.1.0

plainheart (approved):

